PR #17, opened 2017-07-09
Add flag similarity field to Sierra Leone and Galápagos.

ruricolist:
They have nearly identical flags.


axelboc:
So they are, thanks! However, it seems that the difference is in the shade of blue, not the width (they have the exact same ratio). Can you please tweak the similarity information accordingly, and also use the full country name for the Galapagos:

- Galápagos Islands (slightly darker blue)
- Sierra Leone (slightly lighter blue)

Thanks a lot!

ruricolist:
I've updated the pull request with the correct information.

axelboc:
Awesome, thank you!

PR #43, opened 2017-12-03
Add capital hints for Basseterre and Basse-Terre.

ruricolist:
They're not exactly the same, but close enough for a hint to be useful.


axelboc:
Good thinking. While you're at it, would you mind adding hints for **George Town** (_Cayman Islands_) and **Georgetown** (_Guyana_) as well? Perhaps "Caribbean" and "South America", not sure.

code-hunger:
> Good thinking. While you're at it, would you mind adding hints for George Town (Cayman Islands) and Georgetown (Guyana) as well? Perhaps "Caribbean" and "South America", not sure.

These names are hell to learn (along with the Basse[-]terre ones)! I'm still getting them wrong after several months of deck usage (and, to be honest, not paying much of attention).  
Just to complete the group, there is also a *Saint George's*, which adds up with confusion.

Another set of such simmilar names are the `J+n`-like  capitals, again in the countries around the carribeans: Saint John's, San Juan and San José.

Though now I see a very intuitive way to remember the george towns - the _two_-word state (Cayman Islands) has a _two_-word capital (George Town), and the _one_-word state (Guyana) has _one_-word capital (Georgetown).

axelboc:
Haha yeah, they're hard. Your George(-)town mnemonic is awesome, thanks! You could argue that Saint/San is a good hint on its own, but Juan/José is indeed quite confusing. It's also difficult not to give away too much in the hint. I wasn't sure about "Part of France", but can't think of a better alternative.

axelboc:
Nice, what about simply "Not a sovereign country" for _Guadeloupe_? Might be the easiest.

PR #76, opened 2018-11-05
Update german name for Swaziland

SuppSupp:
I think the new german version should also be using the new official name. If you disagree, feel free to just close this PR.


axelboc:
From what I see, there's been some [long discussions](https://de.wikipedia.org/wiki/Diskussion:Swasiland) about it on Wikipedia. Apparently, they've decided to stick with **Swasiland** for now.

Staying true to Wikipedia has been a tremendous help in maintaining the deck so far, so I'm closing the PR without merging. Thanks anyway, @SuppSupp! Please do open a new PR if you notice a change on Wikipedia.

PR #83, opened 2019-03-17
Fix typo (Carpenteria -> Carpentaria)

andreaswachowski:
According to Wikipedia, the "Gulf of Carpenteria" should be the "Gulf of Carpentaria".

(Disclaimer: After reviewing other commits, I have used plain vim and git to fix this. I hope that's sufficient for such a small change, and the Deck Manager is not necessary here.)


axelboc:
Well spotted, thanks!

PR #103, opened 2019-07-20
Add license of some islands and water bodies

aplaice:
See #2.

(As mentioned there, I'm not sure whether keeping the license and attribution information in an unstructured form in `LICENSE.md` is practical, and whether, say, having a `sources.csv` file with three or four fields for media filename, license, source and (perhaps) modifications, might not be better. GitHub seems to have a relatively nice CSV viewer, so such a file could even be quite readable out-of-the-box. Obviously, this is just a suggestion.)


aplaice:
As discussed, this is superseded by https://github.com/axelboc/anki-ultimate-geography/pull/108

PR #104, opened 2019-07-20
Correct the extension of ug-map-saint_helena

aplaice:
As determined by `file` it actually is a PNG, not a JPEG. AFAICT
currently neither Anki nor Ankidroid has issues with the file, despite
the wrong extension, but it might still be worth changing.

(In `LICENSE.md`, the file is actually referred to as
`ug-map-saint_helena.png`.)


axelboc:
Well spotted!! 👓

PR #109, opened 2019-07-22
Improve some of the maps that have slightly better upstream versions

aplaice:
This probably should be split into smaller, atomic commits, but ideally the commits should also update sources.csv, which doesn't yet exist in master. I don't want to continue piling pulls on top of each other, so I'm just pushing this, in case you (or anybody else) has some comments.

The label-free version of the Pacific Ocean on Wikimedia is, for some reason huge (98 kB) and playing around with pngout and optipng doesn't help.

The new sources for the files (for my future reference, when updating `sources.csv`), haphazardly listed are:

- https://commons.wikimedia.org/wiki/File:Arctic_Ocean.png
- https://commons.wikimedia.org/wiki/File:Atlantic_Ocean.png
- Indian Ocean: same file but use PNG.
- Azores: same file but cropped, converted to PNG and optimised
- Celtic Sea: same file but use original PNG (optimised)
- Ceuta: same file but use PNG (407px), optimise
- Denmark strait: same file but use original PNG (optimised)
- https://commons.wikimedia.org/wiki/File:GulfofMexico3D.jpg (note that the original NOAA image was indeed a JPG) (add question mark)
- Melilla: same file but use PNG (407px), optimise
- Polynesia: same source but new version; remove circles and labels, optimise


axelboc:
I've merged two of your PRs so far then rebased this one onto master and force-pushed it (so you may want to do a full checkout or reset rather than a pull). Great stuff, thanks heaps!

Let me know when you're done with this one. I've had a look at it on my end and only spotted one thing: the new `ug-map-indian_ocean.png` has a transparent background whereas the other ocean maps have white backgrounds. Might be slightly smaller in size and more consistent with a white background too.

aplaice:
Thanks for looking at and merging the other PRs.

> I've had a look at it on my end and only spotted one thing: the new ug-map-indian_ocean.png has a transparent background whereas the other ocean maps have white backgrounds. Might be slightly smaller in size and more consistent with a white background too.

Thanks for spotting that! I'll fix the transparency asap.

aplaice:
I've replaced the Indian Ocean map with an older version that has a white background (to save myself playing around with transparencies). It's otherwise identical to the newest version, so I think there should be no issues. As a bonus, looking at older versions made me realise that there's an older, smaller version of the Pacific Ocean map (the addition of the transparency somehow increased the size sevenfold, when it should have increased it by several bytes, at most...), so I used that.

I split the pull request into several commits, as it made it easier to check that I had correctly updated all the necessary files.

axelboc:
Awesome, thanks for the separate commits and for listing the sources 👍 

Would you mind just shrinking the _Denmark Strait_ map down to 500px wide like to match most of the other maps? It looks a bit huge in Anki. The _Celtic Sea_ map is also quite big. Perhaps a height of 500px would do as well?

axelboc:
Oh and any reason why you picked 407px-wide versions for some of the other maps and not 500px-wide?

aplaice:
> Would you mind just shrinking the Denmark Strait map down to 500px wide to match most of the other maps? It looks a bit huge in Anki. The Celtic Sea map is also quite big. Perhaps a height of 500px would do as well?

Yes, of course, I'll do that.

> Oh and any reason why you picked 407px-wide versions for some of the other maps and not 500px-wide?

I picked a width of 407px, as that was the width of the previous JPG versions, but I can standardise it to 500px-wide.

aplaice:
I think that I've fixed all the issues and rebased on top of master.

Slightly experimentally I've reduced the number of colours used for some of the "flat" maps, to reduce file sizes (without this, the "scaled down" maps of the Celtic Sea and Denmark Strait would have been far larger than the original PNGs).

axelboc:
Brilliant, thank you!!

PR #110, opened 2019-07-25
Replace some PNG flags with their SVG versions

aplaice:
Ideally all the remaing PNG flags would be replaced with SVG versions, but some of the SVG's on wikimedia are rather large.  The criterion for choosing which to replace was if the SVG optimised with svgo was smaller than 62 kB (cut-off was raised from my initial goal of 60 kB, so that Saint Helena at 61 kB could just scrape in...).

(I haven't forgotten about providing sources for some of the remaining flags and maps!)


axelboc:
I'm not quite sure about this one. It's basically about raising the cut off I chose initially, which seems to have been 40 KB, up to 62 KB. Is it really worth it? The current PNGs look pretty sharp to me. What's your thinking?

aplaice:
> I'm not quite sure about this one. It's basically about raising the cut off I chose initially, which seems to have been 40 KB, up to 62 KB. Is it really worth it? The current PNGs look pretty sharp to me. What's your thinking?

You're probably right that it's not worth it. Some of the flags are a bit blurry when you zoom in,on Ankidroid, but it's not a major issue.

To be honest, before starting, I had hoped that the size of the SVGs on Wikimedia had been brought down in the meantime (for comparison, see the flag of Portugal, which has quite a lot of detail, and whose size had been reduced by 80 % between 2010 and 2014, to just 13 kB) and that svgo's efficiency had been increased, but neither of these appears to have occurred (not that I'm really allowed to complain since I didn't help in either of these tasks :p).

OTOH considering that ~ 70 of the map files are over 40 kB (21 over 50 kB, 13 over 60 kB) lifting the cut-off to, say, 50 kB might not be *too* extravagant.

One thing that should be fixed irrespective of what happens to this pull request, though, is Fiji's flag which is currently somehow [expanding beyond its borders](https://raw.githubusercontent.com/axelboc/anki-ultimate-geography/cc7c541e49aaef38de21852e0842d4b12d5753f2/src/media/ug-flag-fiji.png) and which it [shouldn't be doing](https://commons.wikimedia.org/wiki/File:Flag_of_Fiji.svg).

axelboc:
Alright, a 50KB limit seems reasonable. It should increase the deck's size
by less than a MB, shouldn't it?

aplaice:
Including the space regained from the removal of the old PNGs, it should increase the deck's size by ~ 90 kB.

Issue #16, opened 2017-06-13
Is there a path to upgrade from v1.0?


ElChapitan:
Really, I'll take just about any path, even if it is difficult.  I'd prefer not to have to start the process over, even though i'd get a lot of quick wins  :-)

axelboc:
I know the feeling 😞 

I'm afraid there's no solution but to start fresh. Personally, I found it quite satisfying to relearn everything from scratch with nice-looking flags and maps.

ElChapitan:
<sigh>

OK.  Next question.  Do you plan on having another breaking change anytime soon?

axelboc:
Oh no, not at all. @code-hunger is working on adding more interesting snippets of country information, but this won't require any changes to the templates and fields. However, I did have issues with importing the deck in the past (i.e. updating to a minor version, or even just re-importing the same version)... I even lost my progress once. Not sure if it was just me, though: if you have time, please, would you mind importing the latest version, using it for a couple of days, re-importing it and letting me know what happens to your progress?

ElChapitan:
No problem, but I also would like to know... what's the upgrade path?

axelboc:
Hmm well you can either download the apkg file from the [releases page](https://github.com/axelboc/anki-ultimate-geography/releases) and just open it, or you can clone or download this repo and import the _Ultimate_Geography_ folder with the [CrowdAnki](https://github.com/Stvad/CrowdAnki) add-on. (Note that you won't be able to use the add-on's GitHub import feature because of the way I structured the repo.)

axelboc:
FYI, I've just added an [_Upgrading_ section](https://github.com/axelboc/anki-ultimate-geography#upgrading) to the README with some clear steps. This won't work for migrating from v1.0 to v2.x, though, as discussed.

PR #114, opened 2019-07-28
Improve 8 maps of the Lesser Antiles

axelboc:
I've added a zoom box to eight locator maps of Lesser Antilles islands. A good start... but still a lot more to go for issue #16 

I chose to show some of the neighbouring islands in the zoom box, for context. I hope this doesn't make the island(s) in focus too small. Here's one of the updated maps as an example:

ug-map-saint_vincent_and_the_grenadines


aplaice:
These look great!

Having some of the neighbouring islands in the box, for context, is IMO more useful than having an even greater zoom of the focused island(s).

Issue #90, opened 2019-04-30
Country and Capital Card Template Update to Mimic Cloze


ohare93:
Let me start by saying that perhaps this is just an issue I have faced.

I've always had an issue where I would misread "Capital" as "Country" and vice versa, when speeding through the cards. I would either spend a long time staring at the 'Capital' in complete bewilderment trying to work out it's supposed 'Country' before realising my mistake, or I would mistake similar looking/sounding Countries and Capitals like "Belarus" and "Belgrade" and give the wrong answer because of this. I was always annoyed by this because *I did genuinely know the answer* but I felt bad about this mistake and marking the answer as Good nonetheless.

So I changed my Card Templates to mimic a cloze deletion.

>Country -> Capital
![image](https://user-images.githubusercontent.com/1471350/56987052-b05d2a80-6b8c-11e9-91ae-35377036fb96.png)

>Capital -> Country
![image](https://user-images.githubusercontent.com/1471350/56987055-b3f0b180-6b8c-11e9-82da-94404cdaa079.png)

This way the Countries are always on the top and the capitals are always on the bottom. This got rid of this issue *immediately* and I believe it has helped my recall as a result. You can ignore the colours and size of "Country" and "Capital", as well as any cosmetic change I have made as a part of this. The style of a cloze deletion is all I am really advocating for. The "???" can also be changed to directly copy the style of a cloze "[...]" in blue.

##### Sidenote: the "Show Map" button is just made by the `{{#Map}}<br>{{hint:Map}}{{/Map}}` and I only rarely click it when I get a new card which I have never heard of, do not know the location, and am not getting the Map card today. Rather than clicking edit and looking at the map, which I used to do a lot (which is seemingly impossible in AnkiDroid). This helps better associate it in my mind with it's neighbours and learn faster.

axelboc:
Ooh I quite like this. It keeps tripping me up too. The fact that the `COUNTRY` and `CAPITAL` headings kind of look the same doesn't help either...

Did you do the same for maps and flags -- i.e. put the country at the top? The difficulty with the _Country - Map_ template is that not all notes are countries (water bodies, continents, regions, etc.)

ohare93:
###### I'm not the only one! 😄 

[...]

PR #118, opened 2019-08-04
Cloze-like Style Card Templates

ohare93:
I have finally found the time to get the Github project setup for this wonderful project, and finalize my Cloze-like style templates for all card types. I have made all card types have the same Cloze-like style, to maintain a consistent look. I also removed some useless parts of the styling, trying to consolidate it a bit.

Here's a preview of each card type:

image

image

image

image

image

image

Also in the process I upgraded to the Extended deck type (though I'll be going through those cards slowly). One thing I do not understand (so I may have done it wrong) is why there is a different Card Type for "Country - Capital EXT"? As all cards are now cloze-like, in this build, this one seems useless, but I have left it in for now.

I hope this is enough to satisfy Issue #90. Thanks again for the great deck! Happy to contribute in the future.


axelboc:
As mentioned in the issue, I'm reluctant to make the change, specifically for two reasons:

- I'm concerned that making a change to the templates will lead to people losing their progress when upgrading the deck with CrowdAnki. It's happened to me before and it's not pretty... However, this is relatively easy to test in a fresh Anki profile, which I'll do.
- More importantly, only you and Erim24 have shown interest in this so far. Even on the comments feed of the [shared deck's page](https://ankiweb.net/shared/info/2109889812), I don't recall anyone reporting having trouble with the current templates or suggesting a switch to cloze-style templates. It doesn't mean it's not a good idea, far from it, but I would have really liked to see more support for it before making such a significant change. It would suck to switch to cloze-style templates only for people to start complaining about them and saying it was better before! :confounded:

That being said, I have yet to give these templates a try ... and I will, promise! :smile:

---

Regarding the `EXT` templates, their purpose was to clarify what had to be guessed for a given country (capital/flag/map) when reviewing a card from the extended deck. The standard deck comes with a single template with the name of the country on the front: the _Country - Capital_ template, so showing the word `CAPITAL` on the front side was unnecessary. This distinction is no longer relevant with your cloze-style templates.

ohare93:
Totally understandable, and what I expected. Since I feel strongly that this is an improvement for these flashcards, I thought I would put in the effort of making the change itself. It still must be discussed further and tested, that is expected. But making the pull request makes it easier to discuss the exact changes proposed, and allows people to try it out too. Whatever is eventually decided, s'no big deal 👍 

Also, if it helps, I overwrote my own cards with different versions of this layout several times, and I lost no progress.

axelboc:
I've had a play with your cloze-style templates @ohare93. Here my thoughts so far:

- I can really see the benefit for the capital templates. It's just so much clearer!
- One (very minor) downside is that the eyes no longer know where to look for the question when the front of a card appears.
- I like that the extended deck's templates are more consistent with the standard deck's templates than they are now (but we could solve this by providing custom templates for the extended deck instead of sharing the standard deck's templates).
- I'm not convinced by the map and flag templates. Having the country with `???` underneath the flag/map is incoherent with the capital templates. I feel like it should go above, but this could become problematic on a phone in landscape mode, for instance, as the image might not be entirely visible right away. If the flag similarities end up below the flags, they could also be missed.
- In the current templates, I had managed to avoid associating the term "Country" with a water body or a continent. In the new templates, this is not the case. However, I think I found a way to show the correct type using the tags and some extra HTML and CSS. Alternatively, we could not show the "Country" type at all--only the "Capital", "Location" or "Flag" type. This would also save space if we chose to put the country above the flag/map.
- Very minor again: I'm not a huge fan of the styling of the hints (colour and size), and I feel like one question mark would be sufficient. Easily adjustable anyway.

I've tried to make a version of the templates where the "Country" type is not explicitly shown, and where the country always appears at the top, even on the flag and map templates. Here are the zips, let me know what you think:

- [Ultimate Geography.zip](https://github.com/axelboc/anki-ultimate-geography/files/3498821/Ultimate.Geography.zip)
- [Ultimate Geography [Extended].zip](https://github.com/axelboc/anki-ultimate-geography/files/3498820/Ultimate.Geography.Extended.zip)

ohare93:
Very indepth look, thank you! Thanks for the deck share, it was useful to take a look. I'm glad you can see the benefits, and I generally agree with everything you said. Your concerns are valid, and I passed my mind too. Now, specifics:

**Country Top/Bottom:** As you said, having Country be on bottom can be confusing at times, but it was in an attempt to mitagate the immediate concern you raised: that info below the screen would be missed. Though with correct application of the `<hr id=answer>` tag (as you have done!) it should be fine in the end. I think the consistency of Country being on top is the right answer 👍 Especially with the removal of "Country" text in general. 

**Styling:** Originally I added the colours to attempt to distinguish Country and Capital, then when even that didn't work I changed it to this Cloze-style. I suppose I just left the colours as I'd gotten used to it. I admit looking at your cards I was initially turned off by how grey and dull they looked, but it quickly lost it's impact and I'm sure I'd get used to it in no time. Also sure, one "?" looks fine. Regardless it is incredibly minor, not the hill I wish to die on! 😅 

**"Country" for Water/Continents/Islands:** I had the same general idea, of using javascript to search the tags for "UG:Oceans+Seas" and replacing the "Country" title with "Ocean/Sea". Which is easy enough to do. But just not showing "Country" at all very much saves us the trouble 💯 👍 I am entirely onboard with that.

Consistency is very important, and while I believe implementing these Cloze-type cards will inherently bring some benefits (more consistency being one of them!) I do agree that your proposed changes to my suggestion are more in line with the deck and bring further consistency. I have no proposed changes to your cards as they stand 🥇 What next then? 😉

axelboc:
Closing for #128 -- thanks for championing this change, @ohare93!

PR #121, opened 2019-08-06
Add sources for the remaining maps

aplaice:
I'm leaving the statement in `LICENSE.md`, about the maps being sourced from Wikimedia  Commons, as the author of the majority of the maps (TUBS) has asked:

> I'd greatly appreciate, that you attribute this media file to Wikimedia Commons, if used outside Wikipedia or Commons.

For example, see here:

https://commons.wikimedia.org/wiki/File:Abkhazia_in_its_region.svg

I think that having the links to specific files in `sources.csv`, in addition, is valuable for consistency and for our own organisational convenience (e.g. checking for newer/better versions).


aplaice:
In light of #120, I'm glad that I had started with the maps. :) 

(I had used a bash script that opened the most likely sources on wikimedia commons, in Firefox, for batches of maps, so that I could visually compare them with the version in `src/media/`, and if there were differences or the most likely source simply didn't exist, I looked up the source manually.)

axelboc:
Haha, looks like we had the same idea. Glad I started with the flags too, then! 😄

PR #122, opened 2019-08-07
Add source for EU flag

aplaice:
Finish off #2.

According to the [source](https://commons.wikimedia.org/wiki/File:Flag_of_Europe.svg), only the implementation of the design, not the design itself, is in the public domain. The copyright for the design lies with the Council of Europe. They allow the use of the flag under the conditions specified [here](https://www.coe.int/en/web/about-us/the-european-flag#column-5).

> Any natural or legal person (‘user') may use the European emblem or any of its elements, subject to the following conditions of use.

> Conditions of use

> The use of the European emblem and/or any of its elements is allowed, irrespective of whether the use is of a non-profit or commercial nature, unless:

> (a) the use creates the incorrect impression or assumption that there is a connection between the user and any of the institutions, bodies, offices, agencies and organs of the European Union or the Council of Europe;

> (b) the use leads the public to believe erroneously that the user benefits from the support, sponsorship, approval or consent of any of the institutions, bodies, offices, agencies and organs of the European Union or the Council of Europe;

> (c) the use is in connection with any objective or activity which is incompatible with the aims and principles of the European Union or of the Council of Europe, or which would be otherwise unlawful.

I've summarised that as "Fair Use", which isn't quite right, as the scope of the license is considerably broader than just "Fair Use".

Arguably, the same issue might hold for some of the other flags that are recent enough not to be automatically in the public domain. The Wikimedia editors addressed that explicitly for some (e.g. Montenegro — [according to Montenegrin laws the flag is public domain](https://commons.wikimedia.org/wiki/File_talk:Flag_of_Montenegro.svg)), but seemingly not for others (e.g. Kosovo).


axelboc:
I'll add a warning to `LICENSE.md` about this. In `sources.csv`, let's keep referencing only the "design implementation" licenses as specified on Wikimedia (so `PD` for the European flag).

If someone decides to misuse the European flag, it's their own responsibility. We, ourselves, use it in accordance with the license, so we're fine.

aplaice:
Makes sense, thanks!

PR #128, opened 2019-08-15
Implement cloze-style templates

### LINKED ISSUE #90: Country and Capital Card Template Update to Mimic Cloze

axelboc:
Fixes #90 - follow-up to PR #118 opened by @ohare93

This PR replaces the current note templates with cloze-style templates, where the thing to be guessed (country, capital, location, flag) is replaced with a question mark.

Contrary to the previous templates, the country (or water body, continent, territory...) now always appears at the top of the card, even when it is the thing to be guessed. This brings consistency across all templates, including the extended deck templates.

I'm now calling on all users and contributors of the deck: @aplaice, @code-hunger, @Erim24, @JohnHBrock, @ruricolist... Can you please try the new templates (zips below) and let me know asap if you have any objections or suggestions? Do they work on your mobiles? Do you like having the country always at the top? ... If everyone's on board, I'll include the new templates in the v3.2 release.

[Ultimate Geography.zip](https://github.com/axelboc/anki-ultimate-geography/files/3505994/Ultimate.Geography.zip)
[Ultimate Geography [Extended].zip](https://github.com/axelboc/anki-ultimate-geography/files/3505995/Ultimate.Geography.Extended.zip)


axelboc:
| Country - Capital |  |
| --- | --- |
| Front | Back |

| Capital - Country |  |
| --- | --- |
| Front | Back |

| Flag - Country |  |
| --- | --- |
| Front | Back |

| Map - Country |  |
| --- | --- |
| Front | Back |

| Country - Flag |  |
| --- | --- |
| Front | Back |

| Country - Map |  |
| --- | --- |
| Front | Back |

code-hunger:
I get the following errors after importing the zip using CrowdAnki - https://pastebin.com/CVpVtT18
I might copy the templates over mine manually, but looking at the pictures I like the new cards :+1: .

code-hunger:
Ok I found in the menus a way to update CrowdAnki, the deck imported successfully now.

With the new design the experience is kind of smoother, because the answer seemingly replaces the question mark, I like that.

I'll check on mobile soon, but the change required collection overwrite, and I'm waiting for it to download now.

+1 from me!

aplaice:
One tiny issue is that on a small screen, for the Map - Country and Flag - Country cards, some additional "scrolling" down and up is now needed to view the image and then the answer, but it's not a huge problem.

The scrolling up part can be solved by adding `id=answer` to the `div` corresponding to the answer. (Unfortunately, I [don't think](https://github.com/dae/anki/blob/c4eca03dd817c254c17334921c113ff0c2a9100b/web/reviewer.js#L83) there's a corresponding `id=question` to save on the initial scrolling down, but, again, it's a tiny issue.)

Erim24:
I added them now to my german deck and will test the cards a few days
Thanks @ohare93 and @axelboc for your work!

JohnHBrock:
I'm pretty new to Anki, what's the best way to test this PR? I downloaded the zip and followed the instructions from the README under **Other languages and versions**, and this PR was installed as a deck called "Ultimate Geography_2" instead of replacing my original "Ultimate Geography" deck. Is that expected? If so, that's fine. I'm curious, though, what would be the right way to modify my original deck directly? Is the usual minor version upgrade process just to import the new version and then delete the old deck?

ohare93:
@JohnHBrock since you're new to Anki I assume you have a very recent version of this deck, which should mean your cards will update after reimporting. Just to double check, which deck did you download and reimport into Anki? You should have taken one of those in [the first comment in this page](https://github.com/axelboc/anki-ultimate-geography/pull/128#issue-307742385), not one from the Releases page, if you wish to test this PR. Also, your cards may have _moved_ to this other deck which was generated, can you double check if this is the case? 👍

axelboc:
Something just popped into my head regarding the _Flag - Country_ and _Capital - Country_ templates: should the _Flag similarity_ and _Capital info_ fields really be displayed under the flag/capital at the bottom of the card, or should they rather be moved under the country so they are more visible when the answer is revealed?

The downside is that it breaks the consistency of displaying an info field next to the item it informs on... If that makes sense? Also, the _Country info_ field means that there could potentially be two consecutive info fields displayed.

ohare93:
My vote would be no. Sure there is a slight worry that people will miss such information, but they can see there is scroll bar, and the use of the `id="answer"` for autoscroll will very much help already. I think you're overthinking it, it's better left as it is with consistency maintained.

axelboc:
Okay, you're right. Merging now, then! I think it's time. We can always roll back if we see that people don't like it.

PR #134, opened 2019-08-23
Improve some maps

aplaice:
I've kept the changes broken up as atomic commits, in order to make further modifications easier, in case further improvements are needed.

Some comments regarding specific maps:

### Azores

The Azores appear further south on the new map than on the old map, because the old map used an unusual projection, which placed the Azores further "up" than points with the same latitude in mainland Europe. (See [the uncropped source of the old image](https://commons.wikimedia.org/wiki/File:Locator_map_of_Azores_in_EU.svg).)

(That's actually the main reason why I decided to switch away from the old map (which was rather nice, if inconsistent in style with most other maps), because, especially in the cropped version we were using, it gave a false impression of the Azores' latitude.)

### Saint Kitts and Nevis

This is closely based on [the map of Antigua and Barbuda with a zoom box](https://commons.wikimedia.org/wiki/File:Antigua_and_Barbuda_in_its_region_(zoomed).svg), which we were already using for Antigua and Barbuda's map. In particular, I've kept the colour scheme (paler blue for the seas than usual), as it must have been a deliberate choice (compare the Commons versions of Antigua and Barbuda [with a zoom box](https://commons.wikimedia.org/wiki/File:Antigua_and_Barbuda_in_its_region_(zoomed).svg) and [without one](https://commons.wikimedia.org/wiki/File:Antigua_and_Barbuda_in_its_region.svg), and generally TUBS's (the author of virtually all our maps) stylistic choices are great.

I can see the idea behind having a paler blue when there's a zoom box present (to help highlight the zoombox), but I'm not sure whether it really works.

You (axelboc) didn't adjust the colour of the seas when [adding zoom boxes](https://github.com/axelboc/anki-ultimate-geography/pull/114), and the three zoom box examples in the "Locator maps design recommendations" section on Wikimedia Commons also don't use a paler than usual blue.

If we decide that the paler blue doesn't work, it would probably make sense to also update our [Antigua and Barbuda map](https://raw.githubusercontent.com/axelboc/anki-ultimate-geography/master/src/media/ug-map-antigua_and_barbuda.png). (I could do that easily.)

### Remaining Lesser Antilles maps

Stylistically, I decided not to follow the example of Antigua and Barbuda (and having all the seas being paler than usual) or of the 8 recently added Lesser Antilles zoomed maps (and having all the seas at the same "normal" darker shade). Instead, I've kept the sea in the "zoom box" paler, but left the sea in the "main view" at its "normal" colour, in order to highlight the zoomed-in view, which I think works well. However, if need be, changing the colour of the seas is trivial. (For the time being I also haven't fiddled with the colours of Saint Kitts and Nevis (above) or of Antigua and Barbuda.)


ohare93:
💯 / 10 change 👌

aplaice:
Just in case I don't have time later and to allow easier comparison, I've updated the colours of the Lesser Antilles maps to follow the same style, as the 8 added in 114 (same "normal" colours both inside and outside the zoom box).

Depending on what consensus regarding colours schemes (see above) develops eventually, this commit (f3f1476) can be either discarded or kept (and ideally rebased/merged into the relevant previous commits, to avoid increasing the size of people's `.git` with unnecessary, never-actually-used binaries (probably only ~ 400 kB, but still)).

For the curious, the sed commands used on the source SVGs:

<details>

```bash
# for the 10 new images
for i in *.svg; do sed -e 's/C6ECFF/B3DFF5/g' -e 's/FEFEE9/FDFBE5/g' -e 's/0978AB/1178AC/g'  -i "$i"; done

# for the one new and one old image
for i in *.svg; do sed
# white
-e 's/FFFCEB/FDFBE5/g' \
# blue \
-e 's/D5EDFC/B3DFF5/g' \
# red \
-e 's/C1202D/C12838/g' \
# globe white \
-e 's/FCFCFD/F8F9FA/g' \
# globe grey \
-e 's/D8D6D6/CECECE/g' \
-e 's/6A6766/646565/g' \
# bounding box grey \
-e 's/6B6867/6A6968/g' \
# border grey \
-e 's/6B6866/656565/g' \
# coastline blue \
-e 's/0374AB/1178AC/g' \
-i "$i";
done
```

(I could indeed update the "upstream" SVGs, at least the ones that I had created myself, but it feels wasteful to upload an extra ~ 50 MB onto Wikimedia's servers, for a change of a couple dozen bytes.)

</details>

axelboc:
Gosh, you rock! 🎸 😄

Seriously, the new maps of Azores, Kaliningrad Oblast and Zanzibar are awesome. The zoom boxes are great too! This is quite subjective, but I'm not sure about the positions of some of them, though:

- For the islands near Puerto Rico, I feel like it would look better centred in the bottom, like on the Antigua-Barbuda map, so it doesn't hide Cuba, Haiti and the Dominican Republic.
- For Trinidad and Tobago, why not put it in the top right-hand corner as for the ABC islands ?

What do you think? Either way, it's a gigantic improvement, so happy to merge them as is, of course.

The case of the colours is a tricky one. I had been kind of burying my head in the sand, hoping the problem would go away... 😂

I was under the impression that the paler blue was the "newer" style of the locator map project, but I'm not sure. I don't know if you've noticed, but the strokes are also thinner. I think that's what makes it feel a bit cleaner to my eyes. Some of the maps have the fuller blue and thinner strokes, so it's all a bit of a jumble... The style of the zoom box probably differs as well,  but I'm not sure in which way. When I added a zoom box to those 8 maps, the base map I used had the fuller blue style, but I didn't notice the difference until I had already made a couple, so I just kept on going.

All in all, I'm not sure what to do. The `sed` command you wrote brings some hope. Perhaps we could normalize all of them automatically. Uploading to Wikimedia, converting to PNG and updating `sources.csv` would remain very tedious... For reference, here is a zip with the 8 raw SVG maps I updated in #114.

[Maps with zoom box.zip](https://github.com/axelboc/anki-ultimate-geography/files/3547802/Maps.with.zoom.box.zip)

aplaice:
> For the islands near Puerto Rico, I feel like it would look better centred in the bottom, like on the Antigua-Barbuda map, so it doesn't hide Cuba, Haiti and the Dominican Republic.
> For Trinidad and Tobago, why not put it in the top right-hand corner as for the ABC islands ?
> What do you think?

Done. (I think you were right — they do look better!)

> I don't know if you've noticed, but the strokes are also thinner.

Actually, I hadn't. :) Adjusting the stroke-width should also be straightforward with `sed` (or a less blunt tool), though, if it proves necessary.

> All in all, I'm not sure what to do. The sed command you wrote brings some hope. Perhaps we could normalize all of them automatically. Uploading to Wikimedia, converting to PNG and updating sources.csv would remain very tedious... For reference, here is a zip with the 8 raw SVG maps I updated in #114.

Since we're using the fuller blue in virtually all the other maps taken from the locator map project, perhaps let's just stick to the fuller blue, for everything, for simplicity? If we end up using more existing maps with zoom boxes, that follow the newer style, then we could revisit this decision. (Slight style differences should be OK. For instance, the style of the mini-map differs a bit for some of the maps — compare Spain and Azerbaijan — but it's not annoying.)

On the whole, possibly your approach of burying your head in the sand (at least until I started stirring up the sand) was most sensible...

axelboc:
Awesome, thanks. You're right, let's stick to the fuller blue. Merging now.

PR #135, opened 2019-08-28
Fix extension in sources.csv

aplaice:
Sorry! (I wasn't expecting #134 to be merged, yet\*, and had been
planning to triple-check everything before then.)

\* though I'm done with improvements in the maps, for the
moment, so it's for the best. :)


axelboc:
Woops, sorry 😄 

I assume so, but are all the PNGs optimised?

aplaice:
Yes, all the PNGs are optimised. :)

PR #147, opened 2019-09-17
Add Norwegian bokmål (nb) translation

### LINKED ISSUE #139: Norwegian (bokmål) translation
Hi! I've got a Norwegian (bokmål; we have two written languages; it's complicated) version of the deck almost ready. I guess I'll just do the same thing as in #119 and add a column to the CSV?

Vages:
Adds Norwegian bokmål (nb) translation to `data.csv` as well as a document explaining translation sources (as this is the first "independent" translation of the deck). Fixes #139

## Testing

I've gone through the entire deck once and corrected all the mistakes I found. I don't think there are many left, if any.

## On the necessity of documenting translation sources

I think it's smart to clarify sources for the translations. Finding out what's right is not always as simple as going to Wikipedia. For example, some Norwegian sources may call the capital of Finland Helsinki, while the foreign ministry recommends the Swedish name, Helsingfors.

Style could also be documented. This may not be the case in all languages, but the Norwegian written language (both bokmål and nynorsk) leaves a surprisingly large number of spelling choices up to the writer. Clarifying style choices helps keep new submissions consistent with the previous (and prevents edit wars).


axelboc:
Impressive work! The translations sources document is a great idea. Also, thanks for adding a section to the contributing guide and for rebasing your branch. 💯

Merging now with commit squashing to keep the history clean. Thanks for all your work! 🏆 

P.S. I'll do my best to release v3.2 this week-end.

Issue #86, opened 2019-04-19
Burundi Capital


beatlesfan087:
The capital of Burundi is now Gitega, not Bujumbura

PR #148, opened 2019-09-21
Add Burundi capital info for English

JohnHBrock:
The Norwegian capital info already seems to say something similar about the capital moving from Bujumbura to Gitega recently, but I just put 2019 instead of February 2019 because I found various conflicting dates on Wikipedia about when the move became official (and the sources pointed to by Wikipedia weren't a help, either).


axelboc:
Original issue for reference: #86

Issue #137, opened 2019-09-01
RFC: Criteria for what items are significant enough to include in the deck?


JohnHBrock:
The deck as it stands is impressively comprehensive, but as I think about what other cards could be added, I wonder where to draw the line. For example, I'd love to include Canadian provinces or maybe the individual islands of Hawaii, but if we start adding sub-national items like provinces/states/departments (or parts of states, in this case of the Hawaiian islands) for every country in the world, the deck will grow very large.

Maybe the rule should be that we only include provinces/states/departments/other subregions if they're exclaves/enclaves or semi-exclaves/enclaves. This seems consistent with the current deck, e.g., French Guiana, Ceuta, Melilla, the Caribbean Netherlands, etc. If this were the rule, then, for example, we would have cards for Hawaii and Alaska, but not cards for individual Hawaiian islands or for British Columbia (since British Columbia is part of the Canadian mainland).

I have a similar question for bodies of water. I want to start contributing various lakes and rivers to the deck, but I'm wondering what the criteria should be for what's OK to include. For example, Lake Superior and Lake Victoria seem like obvious candidates. But what about the much smaller Lake Tahoe and Lake Balaton? Should we include Lake Victoria, but not Lake Malawi? Maybe volume or surface area should be criteria? The deck currently includes the relatively small Sea of Galilee and Dead Sea, but maybe small is OK if there's historic or religious significance?

So what thoughts do you all have on guidelines for what to contribute, @axelboc in particular?

Yorwba:
Regarding sub-national items: I wrote a Python script to generate decks for exactly that purpose, and this issue prompted me to open-source it as [`anki-wikidata-geography`](https://github.com/Yorwba/anki-wikidata-geography). I don't include information like capitals of flags yet, but that shouldn't be hard to add if there's demand (except for the fact that those don't always exist at the sub-national level). Bodies of water might also be possible to support, depending on whether that information is easily available from Wikidata.

axelboc:
Cool stuff @Yorwba!

It's a tough one, @JohnHBrock. Honestly, the current notes of non-sovereign states (Guadeloupe, Java, Sardinia, etc.) are just a legacy of the original Ultimate Geography deck.

I don't think adding internal divisions is plausible. We can't possibly add every country's internal divisions (and which divisions do we add? France has regions and departments, for instance), so we'd have to decide for which countries to add internal divisions, which is way too subjective and biaised.

[...]

PR #149, opened 2019-09-21
Add Saint Martin and Saint Barthélemy

aplaice:
This is for completeness of the Lesser Antilles.

I'm not including the [unofficial flags](https://en.wikipedia.org/wiki/Flags_of_the_regions_of_France#Unofficial_and_traditional_flags_2), following the example of Guadeloupe and Martinique.

In the "Country Info" section, I'm describing them with the generic term "territory" rather than "collectivity", since that is what's already used for French Polynesia (whose political status is similar).

For the German name of Saint Barthélemy, I'm using "Saint-Barthélemy" rather than "Sankt Bartholomäus" since that's what's used in the German Wikipedia article.

I've avoided non-ascii characters (é) in the filenames, though it probably wouldn't have been an issue on any modern filesystem.


aplaice:
(See #137.)

axelboc:
We'll see what happens with #137, but for now it makes sense 👍

aplaice:
I had been thinking about closing this after reading #137, but merging for now also works! :)

Issue #152, opened 2019-10-08
Visual answer type hints


Vages:
I may be a bit dim, but I often miss what kind of answer is expected if I don't concentrate on reading the headline. I guess I'm not alone. Some visual indicator of the type of answer would be great.

An icon in front of the expected type of answer is an idea which I'll deliver a PR on shortly.

axelboc:
That's a cool idea!

One concern would be about emojis not being available on all devices. Not sure that's still the case, these days, though... They also don't look the same everywhere. Looking at your PR, I feel like we could just as well use inline SVG icons in the templates. It would give us more freedom and guarantee the same rendering on all devices.

There are plenty of open source SVG icon/emoji libraries out there. I've just come across https://openmoji.org for instance, which has a few interesting emojis, like the [authority building](https://openmoji.org/library/#emoji=E093), or the [world map](https://openmoji.org/library/#search=map&emoji=1F5FA). They may be a bit too detailed for the size at which they will be displayed in the deck, though.

axelboc:
I've added the `votes wanted` label to the issue. How does everyone else feel about the idea of having emojis/icons in the templates (in whichever format)?

aplaice:
I'd vote for some small (visually and ideally memory-wise) SVG icons!

I feel that emojis are a bit too inconsistent between platforms (like @axelboc pointed out).

> inline SVG icons

Just to clarify, do you mean having the SVG "code" entirely included (inline) in the HTML/templates, (as opposed to just using the standard `<img src="foobar.svg" />` tag, with the actual SVGs bundled as part of the media in the deck)? If yes, I'm not sure if that'd be optimal:

1. Testing ~ a year ago inlined SVGs were somehow mangled by Ankidroid. (This might have changed).

2. It'd lead to messier diffs for the templates.

(PS Thanks @vages for the idea!)

axelboc:
Arf damn, I was hoping inline SVGs were supported. SVG files bundled as media in the deck, sounds good, then. 👍

ohare93:
I've been using @Vages's PR for the last week or so. At the start I did not like the Emojis in the cards, I thought they looked bad; now I simply do not notice them at all :slightly_frowning_face: Also the lack of an icon for the X -> Country cards bothers me aesthetically :slightly_smiling_face: :anger: 

[...]

PR #153, opened 2019-10-08
Add emoji to templates as an answer type hint

Vages:
To make it easier to see which type of answer is expected, I added emoji in front of the question marks. With one exception: Countries; they're such an abstract notion that it's hard to avoid something which conflicts with the other emoji.

The hint emoji I chose are:

- **Capital**: 🏙️ [Cityscape](https://emojipedia.org/cityscape/). Also considered other emoji of houses and buildings, but I think this one is more likely to give the right idea. Not all capitals look like this, but what are you gonna do.
- **Location**: 📍 [Round pushpin](https://emojipedia.org/round-pushpin/). Chosen because the map pin is associated with the concept of location in map software. Also considered the [world map](https://emojipedia.org/world-map/) emoji, but it was to similar to the cityscape emoji on Apple platforms.
- **Flag**: 🏳️ [White flag](https://emojipedia.org/waving-white-flag/). Also considered a black flag, but I just think of white as more of a blank slate than black.

Hope you like the idea!

Solves #152

### DIFF
--- a/src/templates/Country - Capital.html
+++ b/src/templates/Country - Capital.html
@@ -5,7 +5,7 @@
   <hr>
 
   <div class="type">Capital</div>
-  <div class="value">?</div>
+  <div class="value">🏙️?</div>
 {{/Capital}}
 
 --
--- a/src/templates/Country - Flag.html
+++ b/src/templates/Country - Flag.html
@@ -5,7 +5,7 @@
   <hr>
 
   <div class="type">Flag</div>
-  <div class="value">?</div>
+  <div class="value">🏳️?</div>
 {{/Flag}}
 
 --
--- a/src/templates/Country - Map.html
+++ b/src/templates/Country - Map.html
@@ -5,7 +5,7 @@
   <hr>
 
   <div class="type">Location</div>
-  <div class="value">?</div>
+  <div class="value">📍?</div>
 {{/Map}}
 
 --


ohare93:
I had the same issue a while ago, but the new Cloze style templates fixed it for me. This sounds interesting though, I'll try your changes and see how I like it 👍

axelboc:
Closing this one until a clearer decision is reached in #152. Thanks for the suggestion!

ohare93:
As I've been doing more of the extended deck I find I need something that accomplishes the goal of this PR more and more. It is sometimes very difficult to tell which question is being asked, if you are to answer Capital, Flag, or Location. Though I still maintain that just an emoji did not solve the issue (for me, at least). So I made the following updates to my own extended templates:

### Country -> Flag
image

### Country -> Map
image

The big obvious shapes for the flag or the globe really help me get into the right headspace quickly, I am finding at least. The circle for the globe then turning into a square shaped map is asthetically displeasing, but I could not think of a better blank symbol for the Country -> Map, and was unwilling to add in an unreferenced (in a card) media file just to make it look nicer. 

What do you think @Vages, @axelboc? 😁

Vages:
I think this is better than my previous solution, @ohare93. I still want something like this very badly, but before we merge your PR, we should give the community some time to improve upon the idea – there might be someone out there whose imagination is sparked by this. Great work!

PR #162, opened 2019-10-26
Capital -> Country Small Template change

ohare93:
###### [Very small tiny insignificant change]

image

I noticed recently that the "Capital hint" field is removed from the card when the back is shown. As far as I can see Capital -> Country is the only card type which removes info on the back. As a rule, Cloze templates do not remove data, the reveal it. Or so I believe :sweat_smile: 

I personally would like it to stay :smile: 

Interesting side note: I thought "what if the Capital Hint and the Capital Info fields get mixed up to the user?" Well *there are no cards with both a Capital Hint and a Capital Info* as of now :open_mouth: but regardless, if there was it'd look like this:

Screenshot_20191026_080451

The div separates them well enough, in my opinion :smile:

### DIFF
--- a/src/templates/Capital - Country.html
+++ b/src/templates/Capital - Country.html
@@ -17,4 +17,5 @@
 
 <div class="type">Capital</div>
 <div class="value">{{Capital}}</div>
+{{#Capital hint}}<div class="info">Hint: {{Capital hint}}</div>{{/Capital hint}}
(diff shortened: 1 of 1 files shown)


axelboc:
Fine by me, thanks! 👍  

I'll merge after the French translation, just to simplify things... 😄

PR #163, opened 2019-11-05
Norwegian mistranslations and typos 2019 nov

### LINKED ISSUE #160: Cayman Islands Capital hint in Norwegian (bokmål) potentially misleading
I'm not sure about this, as my knowledge of Norwegian is non-existent, so I can only rely on similarities with other languages and Google translate, but the capital hint for the Cayman Islands is potentially misleading.

The capitals of the Cayman Islands and Guyana are very similar, (also in Norwegian), so a capital hint is useful to distinguish them in the Capital → Country cards. However, in Norwegian both countries are apparently\* single words (`Caymanøyene` and `Guyana`). The Norwegian respective "capital hint" fields are:

    Land med to ord
    Landsavn på ett ord

which Google translates to:

    Land in two words
    Country name in one word

and which is potentially confusing, under the circumstances.

\* though I might be wrong — word-splitting rules might be different in Norwegian

(Noticed while reviewing the French translation.)

Vages:
Fixes #160

The capital hint for Cayman Islands and Guyana may be directly transferrable to the English version, if that's of interest (whether the country is part of the mainland or an island).


aplaice:
Thanks!

Sorry to be nitpicky (and again, this might work differently in Norwegian than in the literal translation), but "on an island" suggests a single island, which the Caymans obviously aren't and which might be confusing to the user.

OTOH explicitly stating multiple islands would narrow things down considerably, arguably making the hint overly helpful.

In English I'd use "an island nation", which can refer to any number of islands. Not sure if there's a good Norwegian equivalent.

> may be directly transferrable to the English version

Hmm, might be nice, though the current hints also work well. :)

Vages:
Nitpicking makes things better. I could change it to "I en øygruppe", which means "In a group of islands". I don't think that gives it away; most island nations are groups of islands anyway. Does that sound good to you?

Vages:
It sounded so much like an improvement to me that I just did it. If someone disagrees, I can revert the commit.

aplaice:
> Nitpicking makes things better.

That's why I do it :) (but it can be felt as annoying).

>  I don't think that gives it away; most island nations are groups of islands anyway. Does that sound good to you?

I think you're right. It sounds great! Thanks!

Issue #170, opened 2019-11-16
German capital of San Marino should be Stadt San Marino


axelboc:
... instead of just _San Marino_ - cf. https://de.wikipedia.org/wiki/Stadt_San_Marino

Issue #169, opened 2019-11-16
Greenland and the Faroe Islands are constituent countries


axelboc:
The info boxes of [Greenland](https://en.wikipedia.org/wiki/Greenland) and the [Faroe Islands](https://en.wikipedia.org/wiki/Faroe_Islands) qualify the two states as "Constituent countr[ies] in the Kingdom of Denmark". The _Country info_ field, which currently has them as "Self-governing territor[ies] of Denmark", should therefore be updated.

Issue #168, opened 2019-11-15
Change capitalisation of eSwatini to Eswatini


axelboc:
The [Wikipedia page](https://en.wikipedia.org/wiki/Eswatini) seems to now be going with the capital `E`.

Issue #166, opened 2019-11-15
Capital of Kiribati


josealberto4444:
Wikipedia says that the capital of Kiribati is Tarawa (comprising North Tarawa and South Tarawa), and this deck says it is South Tarawa. Is it an error or there is a reason for that?

aplaice:
Thanks for bringing this up! I think you're right that this should be changed.

The reason why AUG says it's South Tarawa, is that Wikipedia used to give the capital as "South Tarawa", in the infobox, [until 2015](https://en.wikipedia.org/w/index.php?title=Kiribati&type=revision&diff=650372126&oldid=649678202), and AUG is at least that old. (In the leading text, the Wikipedia article *still* mentions the capital as South Tarawa... It also doesn't help that the current articles for both [South Tarawa](https://en.wikipedia.org/wiki/South_Tarawa) and [Tarawa](https://en.wikipedia.org/wiki/Tarawa) claim that their respective subjects are the capital of Kiribati.)

The reason for the disagreement in what the capital is, is that apparently most of the governmental offices are located in South Tarawa.

However, given that our policy here generally is to follow Wikipedia's infobox, where applicable, and given that [Kiribati's governmental (?) website agrees](http://www.kiribatitourism.gov.ki/kiribati-pacific-ocean-location/), it would make sense to change the capital to just Tarawa, possibly with some capital info to the effect of most of the offices being on South Tarawa.

The German and French Wikipedia articles list the capital as "South Tarawa" and "Tarawa-Sud" respectively, so these language versions of the deck should probably still keep/use "South Tarawa" (and its French equivalent) as the capital, though.

josealberto4444:
Thanks for all the clarifications!

> The German and French Wikipedia articles list the capital as "South Tarawa" and "Tarawa-Sud" respectively, so these language versions of the deck should probably still keep/use "South Tarawa" (and its French equivalent) as the capital, though.

I disagree. I think it should be changed in Wikipedia, indeed, so I just changed it and added the references.

aplaice:
> I think it should be changed in Wikipedia

On second thought, you're right.

Some French-language sources that agree with you:

https://www.diplomatie.gouv.fr/fr/dossiers-pays/kiribati/presentation-de-kiribati/

https://publications.europa.eu/code/fr/fr-5000500.htm

I've adjusted the French article to point to these, as well. (Also, some of
the copied references in the French article hadn't quite worked, as in the
[...]

PR #172, opened 2019-11-17
Content updates x4

axelboc:
Resolving four issues:

- #166 Capital of Kiribati (EN)
- #168 Change capitalisation of eSwatini to Eswatini (EN)
- #169 Greenland and the Faroe Islands are constituent countries (EN, DE, NB)
- #170 German capital of San Marino should be Stadt San Marino (DE)

For #169, I've tried to update the German and Norwegian translations. @Erim24 and @Vages, please let me know what they're worth.


aplaice:
Sorry for getting around to looking at this only now, but I'm not sure about the German version of the Country info for Greenland and Faroe Islands. (Note that my German is pretty poor, so take with a grain of salt.)

Firstly, I'm 95 % sure that grammatically, it should be "Landesteil des Königreich**s** Dänemark" (the genitive of "Königreich").

Secondly, (though here I'm getting into even more uncertain territory) I have doubts whether "Landesteil" is the most accurate, German term for Greenland and the Faroe Islands. (To the extent that I can tell, I don't think that it's strictly wrong, and probably not worse than what it replaced, but possibly not the way a German would describe them.)

Landesteil is indeed what is used to describe the given entity in the German articles for [Wales](https://de.wikipedia.org/wiki/Wales) or [Scotland](https://de.wikipedia.org/wiki/Schottland). However, in the German articles for [Greenland](https://de.wikipedia.org/wiki/Gr%C3%B6nland) and [the Faroe Islands](https://de.wikipedia.org/wiki/F%C3%A4r%C3%B6er) the term isn't used at all. In the lead section, Greenland is introduced as "ein autonomer Bestandteil des Königreichs Dänemark". The Faroe Islanders are described as "eine „gleichberechtigte Nation“ innerhalb des Königreichs Dänemark" (just like the Greenlanders).

In the navigation box, at the bottom of both articles, the Faroe Islands and Greenlands are categorised as "Autonome Bestandteile der Reichsgemeinschaft des Königreichs Dänemark" (in the plural).

Hence, in the absence of recommendations from an actual native speaker, and following Wikipedia, I'd suggest using "Autonomer Bestandteil der Reichsgemeinschaft des Königreichs Dänemark", possibly skipping the (der Reichsgemeinschaft).

axelboc:
Yeah, I hesitated... I think you're right. And I'm for skipping the _Reichsgemeinschaft_ part, as we don't use the term in the other languages and it's pretty obscure. Do you mind opening a PR to fix it, if you have time?

aplaice:
Done in #185!

Erim24:
Sorry for the long break. I think the variant with "Autonomer Bestandteil des Königreichs Dänemark" as it is done now is a good way to do it.

Issue #156, opened 2019-10-17
Styling doesn't work well with Night Mode add-on for macOS client


tvanantwerp:
Without Night Mode:
<img width="260" alt="image" src="https://user-images.githubusercontent.com/2487968/67027758-23754180-f0d8-11e9-9bc0-e4966252dff3.png">

With [Night Mode](https://ankiweb.net/shared/info/1496166067):
<img width="262" alt="image" src="https://user-images.githubusercontent.com/2487968/67027786-312ac700-f0d8-11e9-8170-c07379b3667c.png">

This is with Anki for macOS version 2.1.15 and Night Mode plugin version 2.2.4.

ruricolist:
The same problem exists on Windows, again with Anki 2.1.15 and Night Mode 2.2.4.

aplaice:
I can reproduce the issue on Linux, with Anki 2.1.15, Night Mode 2.2.4.

As far as I can tell, there are two issues with the Night Mode styling of `anki-ultimate-geography`:

1. The info text overlaps with the main text.

2. The colour of the info text isn't inverted, so it's difficult to see.

I don't think that either of the issues is really the "fault" of `anki-ultimate-geography`.

<hr/>

To reproduce 1., with the Night Mode addon turned on, but without `anki-ultimate-geography`, (only with a lightly-edited standard card type), change the "Front Template" of a "Basic" card to:

```html
<div id=answer>{{Front}}</div>
<div>overlapping text</div>
```

(obviously remember to keep back-ups of the Template).

To reproduce 2., again with the Night Mode addon, but without the `anki-ultimate-geography` deck, add the following to the "Styling" of a "Basic" card:

```css
div {
 color: #333;
}
```

<hr/>

Looking at the issues filed against [Anki-Night-Mode](https://github.com/krassowski/Anki-Night-Mode/issues/52), I think that issue "2.", above, is the same as https://github.com/krassowski/Anki-Night-Mode/issues/52.

In principle, `anki-ultimate-geography` could use the work-around mentioned there; i.e. we could add something like:

```css
.night_mode .info, .night_mode .type {
  color: #ccc;
}
```

to the CSS.

However, I'm not sure whether it's worth it, and more importantly, it doesn't deal with the more significant problem of the overlap (1.).

(There's also the issue that Anki now has its own in-built night mode, which apparently uses the `nightMode` class, so for completeness, the following would also be needed:

```css
.nightMode .info, .nightMode .type {
[...]

PR #173, opened 2019-11-17
Add styling for Anki's and Anki Mobile's built-in dark mode

aplaice:
This deals with the issue of the info (and "type") text being displayed as dark grey on black.

This is not necessary for Ankidroid, since Ankidroid dark mode automatically inverts all text colours, but doesn't hurt either.

Once #156 is resolved, one could consider adding a selector for `.night_mode`, for the popular dark-mode add-on. (Until then there are more serious (breaking) issues with the combination of ultimate geography with the add-on.)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -42,3 +42,7 @@
 hr {
   margin: 1.5em 0;
 }
+
+.nightMode .info, .nightMode .type {
+  color: #ccc;
+}


ukanuk:
Note on the latest 2.1.20 release, Anki Desktop now recognizes the `night_mode` selector just like on Android. See [this comment on the 2.1.20 beta page](https://anki.tenderapp.com/discussions/beta-testing/1708-anki-2120-beta#comment_48096297) (though it's no longer in beta and is fully released).

Therefore if/when #156 is resolved, instead of adding a selector for `.night_mode` which would be redundant, I'd recommend replacing `.nightMode` with `.night_mode`.

Issue #1, opened 2016-12-28
Review and improve maps


axelboc:
- Check quality and file size of all maps and replace the worst ones.
- Prefer SVG or PNG format rather than JPEG.
- Add source URLs and attributions to LICENCE file.

axelboc:
Experiment with hashes in SVG URLs:

1. Select a few regional SVG maps from Wikipedia
2. Define a path for each country and give it an ID
3. Add `:target { fill: <colour>; }` to the CSS of the Anki cards or of the SVGs
4. Set the `src` attribute of the image in the map field of each note to `<regional-map>.svg#<country-id>`

This could make the deck a lot lighter, as regional maps would not have to be duplicated. It would definitely guarantee consistency across all maps. It may not, however, be supported by the web view that Anki uses.

axelboc:
I replaced most of the old, inconsistent JPEG maps with PNGs: 2be14042088cc874de94604353a47952dbced98e

The new maps are from the German [Locator map](https://de.wikipedia.org/wiki/Wikipedia:Kartenwerkstatt/Positionskarten) project of Wikipedia's [Map workshop](https://en.wikipedia.org/wiki/Wikipedia:Graphics_Lab/Map_workshop). More specifically, they are 500px-wide PNG versions of SVG locator maps from the project's [16:9 regional location map scheme](https://commons.wikimedia.org/wiki/Category:SVG_locator_maps_(16:9_regional_location_map_scheme)). The SVG maps were too big to be included directly, so I downloaded the PNG versions instead and further optimised them with [PNGGauntlet](https://pnggauntlet.com/). Here are a couple of the new maps:

![ug-map-bulgaria](https://cloud.githubusercontent.com/assets/2936402/22314929/de66c45e-e3b7-11e6-859c-1ca5a9d25efc.png)
![ug-map-british_virgin_islands](https://cloud.githubusercontent.com/assets/2936402/22314930/de69d842-e3b7-11e6-9ebc-a11a95378587.png)

Unfortunately, the locator map project is not complete and does not cover some of the maps included in this deck. The maps that **still need improvement** are:

1. The maps of **oceans, seas, straits, bays and gulfs**. I have my eyes on a category called [Locator maps of water bodies](https://commons.wikimedia.org/wiki/Category:Locator_maps_of_water_bodies) which is marked for clean up and may provide a nice, consistent and more exhaustive set of maps in the future.
[...]

Issue #136, opened 2019-08-29
Vectorized ocean maps


NathanaelRea:
I noticed that most ocean and sea maps were low quality, and wanted to see if I could help. The oceanic maps seem to already be on Wikipedia except Polynesia, and so I added them to my fork, which your can see here (still need to adjust colors for consistency):
https://github.com/axelboc/anki-ultimate-geography/compare/master...OldOxygen:vector-maps

I would like to vector all the other ocean maps, but had some questions.
- Currently, the only svg files in the media directory are flags, why is this? Is it because it is suppose to be that way, or that nobody has vectored them yet?
- For each svg file, there is an associated exported png file. Are the vector files required to be exported for some reason?
- The oceanic svg files I added are around 110KB each, which is relatively large. Is this a problem?

axelboc:
Hi there! 👋 Thanks for offering to help out. 🥇 Sorry I took so long to get back to you.

The maps of the sub-regions of Oceania you found on Wikimedia and tweaked to remove labels are amazing, thank you! If you open a PR that I can modify, I'll do the necessary work to convert them to 500px-wide PNGs, optimise them with [TinyPNG](https://tinypng.com/) and add them to the `sources.csv` document (unless you'd like to do all this yourself! 😄). 

---

The reason most maps are in PNG format is for **file size**. I want to keep the deck as small as possible so it can be synced quickly, and stored on low-end smartphones. The limit for SVG files is currently **50 kB** after [SVGO](https://jakearchibald.github.io/svgomg/) optimisation. Anything below this has to be converted to PNG and then optimised again. For this reason, **vectorization is not an end goal in itself for this deck**. Vecotrization is, however, a necessary step to achieve **visual consistency** between all of the maps. Before you jump into it, though, let me fill you in on where we're up to.

The current maps of oceans, seas and water bodies (gulf, bays, lakes, etc.) suffer from two major problems: quality and consistency.

- In terms of quality, the maps are neither accurate nor up to date, and their sources are [mostly unknown](https://github.com/axelboc/anki-ultimate-geography/issues/2#issuecomment-513472371). This means that we can't vectorize them directly. We have to turn to more up-to-date maps on Wikipedia. As an example, nowadays the [Aral Sea](https://en.wikipedia.org/wiki/Aral_Sea) is no way near as big as it is [pictured in the deck](https://github.com/axelboc/anki-ultimate-geography/blob/master/src/media/ug-map-aral_sea.jpg).
[...]

PR #176, opened 2019-11-22
Improve sea maps

aplaice:
See #1.

Data from https://www.naturalearthdata.com/ was used.

The intention was to follow the style of the location map scheme "X in its region" country maps as closely as possible.

Comparison of maps before and after below the fold:

<details>

### ug-map-adriatic_sea.png

before

after

### ug-map-aegean_sea.png

before

after

### ug-map-arabian_sea.png

before

after

### ug-map-baltic_sea.png

before

after

### ug-map-bay_of_bengal.png

before

after

### ug-map-bay_of_biscay.png

before

after

### ug-map-black_sea.png

before

after

### ug-map-caribbean_sea.png

before

after

### ug-map-caspian_sea.png

before

after

### ug-map-denmark_strait.png

before

after

### ug-map-english_channel.png

before

after

### ug-map-gulf_of_california.png

before

after

### ug-map-gulf_of_mexico.png

before

after

### ug-map-hudson_bay.png

before

after

### ug-map-labrador_sea.png

before

after

### ug-map-mediterranean_sea.png

before

after

### ug-map-norwegian_sea.png

before

after

### ug-map-persian_gulf.png

before

after

### ug-map-red_sea.png

before

after

### ug-map-white_sea.png

before

after

</details>

As [already mentioned in the linked issue](https://github.com/axelboc/anki-ultimate-geography/issues/1#issuecomment-554747768), I used comparisons with data from the [1953 IHO publication](https://epic.awi.de/29772/1/IHO1953a.pdf), via visual comparison with the [shapefiles from marineregions.com](http://www.marineregions.org/downloads.php#iho) rendered with `mapshaper-gui`, and by reading the publication itself, in order to ascertain which sub-seas and bays were parts of "parent" seas. I usually double-checked with Wikipedia's pages for the relevant bodies, in case of doubt.

Please point out all and any improvements, regarding style, cropping (choice of surrounding area to be displayed) or any other issues, since I now have an automated workflow, making modifications straightforward.

Ideally and ultimately, I'd like to "upstream" the source SVGs to Wikimedia, though I'll only do that once I've finished improving/modifying the maps, in order to avoid unnecessarily clogging up Wikimedia's servers with several versions of multiple ~ 5 MB files.


ohare93:
Fantastic work mate! :grin: they all look excellent!

axelboc:
OMG - that's all I have to say... 💯

aplaice:
Thanks for the kind words!

In that case, I'll continue with:

1. The "Pacific-adjacent" seas, once I figure out how to re-cut the shapefiles at a different longitude, most cleanly (so that the map can be centred at a different longitude, to avoid the distortion present near the "edges" of the Winkel III projection).

2. The Aral sea, for which I'd like to have its current area superimposed on the historic one. (I think/hope that [Natural Earth Data's "historic lakes"](https://www.naturalearthdata.com/downloads/10m-physical-vectors/10m-lakes/) contains the necessary data.)

mighty-cthulhu:
Great improvements, many fixes. Also I'm glad that the map of the White Sea no longer contains spoilers.

axelboc:
Fixes #183

aplaice:
Some further updates. Comparisons below the fold (to avoid clogging up space):

<details>

### ug-map-coral_sea.png

before

after


### ug-map-east_siberian_sea.png

before

after


### ug-map-gulf_of_carpentaria.png

before

after


### ug-map-gulf_of_thailand.png

before

after


### ug-map-philippine_sea.png

before

after


### ug-map-sea_of_japan.png

before

after


### ug-map-south_china_sea.png

before

after


### ug-map-tasman_sea.png

before

after


### ug-map-yellow_sea.png

before

after

</details>

I'm not quite sure about the East Siberian Sea — the old one, while inconsistent with the style of our remaining maps, is really quite nice.

As before, feedback regarding cropping and zoom are very welcome.

On the topic of the extent of the seas (which areas to highlight) I made the following decisions where the IHO 1953 *Limits of Oceans and Seas* didn't agree with Wikipedia, generally erring in favour of Wikipedia:

1. **Philippine Sea.** According to the IHO 1953 "specification" (and also the latest, still unaccepted draft from 2002) all the inland Philippine seas (e.g. Bohol Sea, Sibuyan Sea etc.) are part of the Philippine Sea. Wikipedia disagrees, unambiguously placing only the Leyte Gulf and Dinagat Sound within the Philippine Sea. Wikipedia is not clear on whether the Davao Gulf belongs to the Philippine Sea, but I included it, anyway as it makes the map look nicer. (For the record, Britannica seems to agree with Wikipedia.)

2. **Sea of Japan.** It's not clear whether the Tsugaru Strait is part of the sea of Japan. IHO says that it is, Wikipedia isn't clear, so I included it.

3. **South China Sea.** IHO (1953) includes all of Taiwan Strait, Gulf of Tonkin, Qiongzhou Strait and Natuna Sea within the South China Sea. IHO (1986 (draft) and 2002 (draft)) don't. Wikipedia hedges, quoting both. I followed the (approved) 1953 version.

As you'll probably notice, the Bering Strait map hasn't been updated as Natural Earth Data doesn't describe it in the relevant shapefile.

Remaining tasks:

1. Aral Sea map (with historical and current extent) — should be relatively straightforward as Natural Earth contains the needed data.

2. (?) Celtic Sea — not in Natural Earth, so will try to pull in from [here](https://doi.pangaea.de/10.1594/PANGAEA.777975) (though my previous experience with this source wasn't great).

3. (?) Bering Strait — not in any of the shapefiles I had looked at until now, so would have to do it from scratch (but a strait with two straight edges should be doable).

4. (?) Oceans — these should be straightforward, but I'm not sure whether they need replacing and what style would be best. Perhaps something analogous to the current continent maps (e.g. [Asia](https://raw.githubusercontent.com/axelboc/anki-ultimate-geography/master/src/media/ug-map-asia.png)) would work?

5. (?) The island areas of the Pacific Ocean. Perhaps since the style would have to be different from everything else we have, anyway (we don't have a style for highlighting areas containing both seas and land) the maps found by @OldOxygen in #136 could be used. ([Polynesia](https://github.com/OldOxygen/anki-ultimate-geography/blob/38d3eb72f79636f9f8b6032b8fc175e000e3ca90/src/media/ug-map-polynesia.svg), [Micronesia](https://github.com/OldOxygen/anki-ultimate-geography/blob/38d3eb72f79636f9f8b6032b8fc175e000e3ca90/src/media/ug-map-micronesia.svg) and [Melanesia](https://github.com/OldOxygen/anki-ultimate-geography/blob/38d3eb72f79636f9f8b6032b8fc175e000e3ca90/src/media/ug-map-melanesia.svg).) The single best "upstream" Wikimedia source seems to be [here](https://commons.wikimedia.org/wiki/File:Oceania_UN_Geoscheme_Regions_with_no_text.svg). Though if we do do that, it should probably be in a separate pull request as it's partially unrelated.

6. Get things upstream to Wikimedia (ideally I should get in contact with somebody at the [Map Workshop](https://en.wikipedia.org/wiki/Wikipedia:Graphics_Lab/Map_workshop) so that I know what norms to follow before dumping tens of huge SVG maps onto Wikimedia...)

axelboc:
Outstanding! 🎉🎉🎉

I feel like the ocean maps are satisfactory for the time being. They have their own style, but they're clean and consistent. So don't worry about them, I reckon 😉

The maps of the Pacific regions would indeed have to be replaced in another PR.

Gosh... Looking forward to merging this. Good luck with the remaining items!

aplaice:
I have no idea why the latest commit failed the checks (and what the third mysterious check actually is...). `src/data.csv` does not indeed follow quote normalisation, but that was also the case for the previous commits, as well as the-then master (and the relevant check is the "index" one, anyway).

As an aside, in case this wasn't a deliberate choice (in order not to scare away new contributors, but have the relevant information somewhere under the hood — though access isn't exactly great), the "index" check will always succeed, irrespective of quote normalisation, since 

    git diff-index --quiet HEAD -- || echo "You need to run \`composer index\`"

will always return an exit code of 0. (If `git diff-index ...` returns a code of 1, then `echo ...` is run, which will always succeed, giving a code of 0.) If having the check fail when `src/data.csv` isn't normalised *was* the intention then perhaps we could have something like:

    git diff-index --quiet HEAD -- || (echo "You need to run \`composer index\`"; exit 1)

aplaice:
Oh, and for easy reference the Aral Sea maps:

<details>

before

after

</details>

The dotted line in Kazakhstan delineates the [Baikonur Cosmodrome](https://en.wikipedia.org/wiki/Baikonur_Cosmodrome) which is an area leased out to Russia. I think it's worth having, (I had deliberately not filtered it out from the Natural Earth dataset, as I had for some others (we don't need *all* the different borders between North and South Korea)) as it's a special, weird sort-of-border, but I'm also open to removing it, if it's undesirable/potentially confusing.

axelboc:
I pushed a commit to fix a syntax issue in the workflow file last week. Perhaps a rebase would make that third check go away?

I wasn't sure how to get the command to fail, so thanks for the tip! But I eventually thought it might as well not, as you suggest.

In regard to the Aral sea map, I do feel the border of the Baikonur Cosmodrome does not need to be there, given the context. My first thought was that it was an old part of the sea, which didn't make sense but was nonetheless confusing.

mighty-cthulhu:
> The dotted line in Kazakhstan delineates the Baikonur Cosmodrome which is an area leased out to Russia.

I found this interesting, but only if this info is added to the card, otherwise it's confusing.

Sadly, Aral Sea looks even drier now in [Google Maps (2019](https://www.google.com/maps/@45.5480462,59.1260697,391714m/data=!3m1!1e3)) than on  Natural Earth map  - I think they last updated it in 2012.

aplaice:
> Perhaps a rebase would make that third check go away?

I'll do a rebase (perhaps not today — probably tomorrow), though why didn't the third check come up after yesterday's? (I'm not expecting an answer :) — if I cared enough I'd e-mail github support as suggested in the check details, but hopefully it's just a one-off issue.)

> I do feel the border of the Baikonur Cosmodrome does not need to be there, given the context.

> I found this interesting, but only if this info is added to the card, otherwise it's confusing.

I don't think the info could be added to the Aral Sea card, so I'll remove the border.

> Sadly, Aral Sea looks even drier now in Google Maps (2019) than on Natural Earth map - I think they last updated it in 2012.

Yeah, you're right, but I don't have any easily accessible, fresher shapefiles. (According to the [changelog](https://github.com/nvkelso/natural-earth-vector/blob/master/CHANGELOG) the latest change is from 2018, and it did touch the lake shapefile, but it's quite likely they didn't update the Aral Sea border.) I think the cost-benefit analysis is definitely in favour of sticking with the existing one than trying to make a new one myself.

mighty-cthulhu:
> I don't think the info could be added to the Aral Sea card, so I'll remove the border.

Remove it if you want. However I thought that "Country Info" field could be used for this info.

axelboc:
> Remove it if you want. I thought that "Country Info" field could be used for this info.

Unfortunately, the _Country info_ field appears on cards that don't have maps (e.g. _Country > Capital_), in which case the information would not make sense.

mighty-cthulhu:
> Unfortunately, the Country info field appears on cards that don't have maps (e.g. Country > Capital), in which case the information would not make sense.

Aral Sea doesn't have a capital or a flag so the only card is Map -> Country.

ohare93:
> Aral Sea doesn't have a capital or a flag

*Yet* 😉

ohare93:
I prefer removing the Baikonur Cosmodrome border info from the map, I don't think it adds anything useful.

axelboc:
Haha true, silly me.

Yeah, still think the information is not relevant anyway.

aplaice:
Rebased and Baikonur Cosmodrome border removed.

<hr/>

I agree with you @mighty-cthulhu that the presence of the Baikonur Cosmodrome border is a fun fact. However, there are many possible interesting facts about the surroundings of the Aral Sea (any one of which might have been useful for remembering the Aral Sea), they can't be all included and there's no objective way of choosing only one of them, so we err on the side of not providing any.

The Baikonur Cosmodrome border could only have been kept on the map if it could stand on its own, without explanation, like the border between China and Hong Kong's new territories (leased from China in 1898) definitely could have until 1997, or like the border between Panama and the Panama Canal Zone (leased out to the USA) arguably could have until 1979/1999, but the consensus here is that it can't.

aplaice:
I'm not sure that the Celtic Sea map should be updated, as the old one is really nice and detailed (though FWIW according to the IHO the south-western limit *is* supposed to be a straight line). For easy comparison see here:

<details>

before

after

</details>

I've also updated the English Channel map using the same source as used for the Celtic Sea map, so that it's slightly more accurate. (The Natural Earth Data was non-existent for the Celtic Sea and slightly wrong for the English Channel, so I used [this](https://doi.pangaea.de/10.1594/PANGAEA.777975). I explained the rationale for not using it for *all* the maps in a slightly over-long commit message.)

<hr/>

I don't think that I'll have time to do the Bering Strait map properly in the next week or so, so I think it would make most sense for me to include it in a later, separate pull request.

I also won't have time to "upstream" the SVGs from which the PNGs are being made, in the very near future, so I'll update `sources.csv` appropriately once I do, in another, future pull request.

Just in case anybody is curious [here's](https://github.com/aplaice/svg_sea_maps) the extremely messy spaghetti-code-like set of scripts for generating both the SVGs and the PNGs. Hopefully nobody other than me will ever need to look at the code.

axelboc:
I much prefer the new Celtic map!

I take it you're ready for me to click the _Merge_ button!?

aplaice:
> I much prefer the new Celtic map!

Great! :)

> I take it you're ready for me to click the Merge button!?

Yes! (I've just briefly re-checked the whole pull-request, just in case, and didn't see any issues.)

PR #178, opened 2019-11-23
fix spelling

horwitz:
"Timer-Leste" -> "Timor-Leste"
"i.e." -> "e.g." (in one instance where it prefaced an example)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,7 @@ A note's _Flag similarity_ field contains a list of countries, each followed by 
 Flag similarities are always **mutual**: if flag A is similar to flag B, then flag B is similar to flag A. To determine whether two flags are similar enough to warrant mutual _Flag similarity_ information, their differences must first be **identified** and **classified**. The following classification applies:
 
 - **Critical differences (C)**
-  - presence/absence of decoration - e.g. symbol, coat of arms, etc.
+  - presence/absence of decoration - i.e. symbol, coat of arms, etc.
 - **Major differences (M)**
   - same colours in different positions (e.g. two swapped, three rotated)
   - decorations of different types in same position (e.g. symbol vs. coat of arms)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ Ready to start working on an issue? Here is what you need to know.
 
 1. Install the [CrowdAnki add-on](https://github.com/Stvad/CrowdAnki) in Anki.
 1. Fork and clone this repository on your machine.
-1. With the help of your favourite command-line package manager (e.g. `brew`, `chocolatey`, `apt-get`), install [PHP 7](http://php.net/) and [Composer](https://getcomposer.org/download/).
+1. With the help of your favourite command-line package manager (e.g., `brew`, `chocolatey`, `apt-get`), install [PHP 7](http://php.net/) and [Composer](https://getcomposer.org/download/).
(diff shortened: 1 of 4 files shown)


ohare93:
Stating what the PR actually does is always helpful :sweat_smile: Helps others get a general sense, rather than having to look themselves. So having a look your change seems to be:
1) Changed all instances of "e.g." into "e.g.,"
2) Fixed some minor spelling mistakes

So about change 1, I thought "dear god that looks hideous! A comma after a full stop :nauseated_face: no one would ever do that! :disappointed_relieved:" was my immediate reaction :smirk:

Then I decided to have a look around and find where this would ever be used, and found my answer in [this wiktionary page on e.g. ](https://en.wiktionary.org/wiki/e.g.) It says:

> "E.g." and its examples are typically set off from the rest of the sentence by punctuation. In US English, "e.g." sometimes will be followed by a comma. However, usually it is not, and it is never followed by a comma in other English-speaking parts.

As a UK native I have literally never seen "e.g.," and it looks quite peculiar to me :sweat_smile: I would vote to not make this particular change in the deck, especially since it is so so minor.

horwitz:
Thanks for clarifying that (and with a source to boot)! I've changed the "e.g.," and "i.e.," back to "e.g." and "i.e." respectively, as per Wiktionary's stance on British English*—admittedly I'm not sure about Wiktionary's general take (since its opinion on US English differs from major sources (like the Chicago Manual of Style and the (otherwise) notoriously comma-stingy AP Style Guide), but I'm for sticking with the status quo since I can't offer a strong reason to change (and as far as I can tell it's the right thing to do—now that @ohare93 cleared things up).

That said, there now remain two small changes in the PR—but I hope that smallness of the changes doesn't mean that the PR should be ignored!

* I'm assuming that this deck uses British English—oddly I hadn't noticed even with all the times I've looked at the cards (but now see "favourite", "colour", etc.)

README.md
```diff
@@ -54,7 +54,7 @@ To help with memorisation and provide context while learning, most notes include
 
 - **similar flags** - e.g. Iceland = _Norway (red background, blue cross)_
 - **governance information** - e.g. Cayman Islands = _Overseas territory of the United Kingdom_
-- **alternative and former country names** - e.g. East Timor = _Also known as Timer-Leste_
+- **alternative and former country names** - e.g. East Timor = _Also known as Timor-Leste_
```
axelboc: Great catch! 👀

CONTRIBUTING.md
```diff
@@ -74,7 +74,7 @@ A note's _Flag similarity_ field contains a list of countries, each followed by
 Flag similarities are always **mutual**: if flag A is similar to flag B, then flag B is similar to flag A. To determine whether two flags are similar enough to warrant mutual _Flag similarity_ information, their differences must first be **identified** and **classified**. The following classification applies:
 
 - **Critical differences (C)**
-  - presence/absence of decoration - i.e. symbol, coat of arms, etc.
+  - presence/absence of decoration - e.g. symbol, coat of arms, etc.
```
axelboc: The intent was to clarify what is meant by "decoration", rather than to give examples of decorations that might be covered by the statement. Although there's really not a huge gap between the two, I think `i.e.` is correct.

axelboc:
I indeed used British English when I cleaned up the deck initially, but I haven't paid a close attention to it since then, so there might be inconsistencies here and there. Don't hesitate to open a PR if you spot any! 😉 The important thing is consistency, so we might as well stick with British English, I reckon.

PR #190, opened 2019-12-02
make hints consistent

### LINKED ISSUE #179: Make hints consistent
Half of the 10 hints area already simply the entity's polity—this issue suggests making all 10 of the same form, specifically:

```
Country               | Capital     | Capital hint (OLD)      | Capital hint (NEW)
----------------------+-------------+-------------------------+---------------------
Guadeloupe            | Basse-Terre | Not a sovereign country | Overseas department
Saint Kitts and Nevis | Basseterre  | Sovereign country       | Sovereign country
Cayman Islands        | George Town | Two-word country name   | Overseas territory
Guyana                | Georgetown  | One-word country name   | Sovereign country
Jamaica               | Kingston    | Caribbean               | Sovereign country
Norfolk Island        | Kingston    | Oceania                 | External territory
England               | London      | Constituent country     | Constituent country
United Kingdom        | London      | Sovereign country       | Sovereign country
Aruba                 | Oranjestad  | Constituent country     | Constituent country
Sint Eustatius        | Oranjestad  | Special municipality    | Special municipality
```

(The table above only covers the English version, of course.)

horwitz:
Fix #179 

Note: changes in the three non-English `Capital hint` fields were done mechanically—in at least one case, that may have led to oversimplification/generalization: for Norfolk Island's `Capital hint:nb`, simply "territory" (in Norwegian) was used. This could definitely use double checking by German/French/Norwegian speakers.

Compare Norfolk Island's `Country info:nb` field, "Selvstyrt australsk territorium."—that seems to mean "self-governing" (which really* shouldn't be in the hint), "Australian" (which shouldn't be in the hint) and, well, "territorium" (territory)—perhaps there's an official** term in Norwegian for "**external** territory"?

*It seems that Norfolk Island was at least **partially** self-governing for a while, but that is no longer the case (which would suggest that the `Country info:nb` field might need to be updated (in a separate PR))
**Per the Australian government?


aplaice:
I'm generally in favour of standardising the hints.

However, I'm not sure whether making the Guadeloupe hint more specific (when it's not necessary) is optimal. If anything I'd move over all the pairs for which it makes sense (i.e. not the Aruba/Sint Eustatius pairs, and nor the UK/England/BIOT triad) to all using "Sovereign Country"/"Not a sovereign country".

The purpose of the hint is to help distinguish two countries with the same (or very similar) capital, while (IMO at least!) providing as little actual help in guessing/remembering the country, as possible.

In my view, an ideal "hint" (in this context) doesn't help you in remembering the name of the country, but once you come up with an answer (though before you reveal the answer card) allows you to validate whether you gave the desired answer for this note, and if not to try remembering again.

(Capital "hint" is IMO a bit of a misnomer.)

Ideally, when you see the capital (e.g. Basseterre) on a card, you remember **both**/all the countries with that capital (in this case, Guadeloupe and Saint Kitts and Nevis), even without looking at the hint, and then "validate" that one of these fits the hint for this given card.

The point is about creating the associations:

{Capital} Basseterre → Guadeloupe

and 

{Capital} Basseterre → Saint Kitts and Nevis

but not the association:

{Capital hint} Overseas department → Guadeloupe

For this purpose, having as few possible capital hints as possible (the pair "Sovereign"/"Not sovereign" being used for almost everything that needs a capital hint) would be useful, as it would help avoid creating an association between the text of the hint and the country rather than between the capital and the country.

(Ideally, every capital would have several possible hints and every time the card came up, one of them would be randomly selected, hence almost completely avoiding association between the capital hint and the country. This is doable with javascript embedded in the cards, but is probably overkill.)

horwitz:
@aplaice 's notes look good to me—thanks!

1. Of the five pairs, what about Oranjestad—neither "country" (Aruba / Sint Eustatius) is a sovereign country (so we presumably can't go the "Sovereign country" / "Not a sovereign country" route)?
2. Instead of "hints" these could be called "disambiguators".

src/data.csv
```diff
@@ -1,5 +1,5 @@
...
-crr.AfnVRi,England,England,Angleterre,England,"Constituent country of the United Kingdom.","Landesteil des Vereinigten Königreichs.","Nation constitutive du Royaume-Uni.","Land som utgjør en del av Storbritannia.",London,London,Londres,London,,,,,"Constituent country",Landesteil,"Nation constitutive","Ikke selvstendig land","<img src=""ug-flag-england.svg"" />",,,,,"<img src=""ug-map-england.png"" />",UG::Europe
+crr.AfnVRi,England,England,Angleterre,England,"Constituent country of the United Kingdom.","Landesteil des Vereinigten Königreichs.","Nation constitutive du Royaume-Uni.","Land som utgjør en del av Storbritannia.",London,London,Londres,London,,,,,"Not a sovereign country","Kein souveränes Land","Pas une nation souveraine","Ikke selvstendig land","<img src=""ug-flag-england.svg"" />",,,,,"<img src=""ug-map-england.png"" />",UG::Europe
```
aplaice: Unfortunately, I think that the German and French can't be changed to "Kein souveränes Land"/"Pas une nation souveraine", here, since in the German and French versions of the deck, the British Indian Ocean Territory is also listed as having London (or Londres) as its capital (and it is also "not a sovereign country").

I think it would be best to keep the current disambiguators in French and German for England.

aplaice:
> 1. Of the five pairs, what about Oranjestad—neither "country" (Aruba / Sint Eustatius) is a sovereign country (so we presumably can't go the "Sovereign country" / "Not a sovereign country" route)?

I'd vote to keep the Aruba / Sint Eustatius hints as they are.

> 2. Instead of "hints" these could be called "disambiguators".

That's definitely a better term! :)

src/data.csv
axelboc: The French hint is fine if we remove the capital and hint for BIOT, but yeah the German hint has to stay as it is. @horwitz do you mind dealing with #192 and reverting the German hint for _England_, please?

axelboc:
Thanks for that!! 🎉

PR #193, opened 2019-12-26
Make style of Bering Strait map consistent with remaining maps

aplaice:
The northern and southern limits of the strait are based on the [IHO's 2002 draft of "Names and Limits of Oceans and Seas"](https://web.archive.org/web/20190529115635/http://www.iho.int/mtg_docs/com_wg/S-23WG/S-23WG_Misc/Draft_2002/S-23_Draft_2002_NORTH_PACIFIC.doc).

However, since the coordinates of Mys Chaplina (64°50'N – 172°20'W) do not agree with those given on Wikipedia ([64°24′17″N 172°13′39″W](https://en.wikipedia.org/wiki/Cape_Chaplino)) or those implied by the map on page 21 of [the aforementioned document](https://web.archive.org/web/20190529115635/http://www.iho.int/mtg_docs/com_wg/S-23WG/S-23WG_Misc/Draft_2002/S-23_Draft_2002_NORTH_PACIFIC.doc), I've adjusted the south-western vertex to the coordinates given by Wikipedia for Mys Chaplina.

For easy comparison, old and new maps:

<details>

before

after

</details>

Also, Merry Christmas/Happy Chanukah/Season's Greetings!

Obviously, please don't feel under any pressure to look at these pull requests now!


axelboc:
Looking great. As always, thanks for the thorough research and documentation. Season's Greetings to you too! 🎉

PR #198, opened 2019-12-27
Fixes in German translation

### LINKED ISSUE #197: [de]Tarawa now capital of Kiribati
The German wikipedia page of Kiribati now changed the capital to Tarawa (like already discussed in #166)
Therefore, I changed it in the German translations and will create a pull request for it.

Erim24:
- Changed "souveränes Land" into "souveräner Staat" as that fits better.
- the german wikipedia updated the capital (South Tarawa to Tarawa): [https://de.wikipedia.org/wiki/Kiribati](url). Fixes #197 
I changed it even though the article about Tarawa says that South Tarawa is the capital, as it was changed on the country article and also in other  #languages(#166).
- fixes a typo in the country info (de) for Kosovo
- adding a capital hint for Aruba
Do you think it should be brought to the same as the country info? There it is a bit more extensive (mentioning that Aruba is part of the Netherlands)


aplaice:
> Changed "souveränes Land" into "souveräner Staat" as that fits better. 

> the german wikipedia updated the capital (South Tarawa to Tarawa): https://de.wikipedia.org/wiki/Kiribati. Fixes #197

These look great! :)

> fixes a typo in the country info (de) for Kosovo

> adding a capital hint for Aruba

Unfortunately, these two seem to have disappeared when you merged `3fbad5d` (your old branch) and `fe897da` (current master).

Aruba has had a capital hint added ("Autonomes Land"), though "Landesteil" may well be better. (I had used "Autonomes Land" based on the country info.)

>  Do you think it should be brought to the same as the country info? There it is a bit more extensive (mentioning that Aruba is part of the Netherlands)

I think that it's best when the capital hints give as little information as possible (while still obviously distinguishing between the countries that have the same capital). @axelboc [seems to agree with this general idea](https://github.com/axelboc/anki-ultimate-geography/issues/179#issuecomment-565804521), so I'd avoid adding the information about Aruba being part of the Netherlands.

Erim24:
I'm still sometimes a bit overwhelmed with git XD
I now updated the typo for the country info of the Kosovo.
Regarding the info about Aruba I just checked a few wikipedia pages and according to them the way we have it now is fine. (I'm unfortunately not familiar enough with the guidelines in this case, but on the description of the Netherlands "Autonomes Land" is used for these islands)

aplaice:
>  I now updated the typo for the country info of the Kosovo.

Great! :)

>  Regarding the info about Aruba I just checked a few wikipedia pages and according to them the way we have it now is fine. (I'm unfortunately not familiar enough with the guidelines in this case, but on the description of the Netherlands "Autonomes Land" is used for these islands)

I agree that that means that we can keep the current hint.

> I'm still sometimes a bit overwhelmed with git XD

Yeah, git can be really annoying. In such situations, where master has diverged considerably from your branch, I'd generally recommend rebasing rather than merging, as it makes it slightly easier to check that your commits are doing what they're supposed to. However, given how unwieldy `data.csv` has become, no solution would be particularly neat.

Issue #161, opened 2019-10-22
Spanish translation


josealberto4444:
I think I could manage to do the Spanish translation of this deck. Is that OK? =)

josealberto4444:
Taking a look at how the translation works, and regarding how git works, I think it's better to wait until the French translation is finally revised and merged. Do you agree?

aplaice:
Thanks for offering!

> Taking a look at how the translation works, and regarding how git works, I think it's better to wait until the French translation is finally revised and merged. Do you agree?

On the whole, yes. (In principle, "merging" the two translated versions would be a matter of copying several columns in a spreadsheet and rebasing, but it's unnecessary additional work and it increases the scope for errors. Unless you have a strong preference for doing the translation now (for whatever reason), it would indeed be better to wait.)

An improved translation system is on the radar (#143), but not yet designed or implemented...

josealberto4444:
I will wait, yes, no problem. =)

josealberto4444:
I will start with this in the following days, as the French translation is already merged. =D

josealberto4444:
I have a question: where are the strings as "Location", "Capital", etc. translated? I cannot find them in `data.csv`.

josealberto4444:
Another question: what do you use for editing `data.csv` creating new columns or things like that?

I was using LibreOffice, but when I save the csv I can only choose to put between quotes all strings or none, while the original file uses quotes only when the string has spaces.

aplaice:
> I have a question: where are the strings as "Location", "Capital", etc. translated? I cannot find them in data.csv.

They aren't currently. (This is ["on the radar"](https://github.com/axelboc/anki-ultimate-geography/issues/143#issuecomment-546508579), but not yet implemented.) 

The strings are only present in the [templates](https://github.com/axelboc/anki-ultimate-geography/tree/master/src/templates) (e.g. [Capital -> Country](https://github.com/axelboc/anki-ultimate-geography/blob/master/src/templates/Capital%20-%20Country.html)) so, in the meantime they could, in principle, be translated by the end-user in their own decks. (Yes, I know that that's horrible and abominably ineffective...)

> I was using LibreOffice, but when I save the csv I can only choose to put between quotes all strings or none, while the original file uses quotes only when the string has spaces.

[...]

PR #201, opened 2020-02-09
Spanish translation

josealberto4444:
Finally, I finished the translation.

If you want me to change anything, I'm open to suggestions, of course. =)


aplaice:
Thank you! :)

A couple of minor issues:

1. For Eswatini, possibly add "Esuatini" as alternative name in the "Country info" (perhaps something approximately like "Oficialmente conocido como Esuatini desde 2018"?)

2. Should the capitalisation of "Mar blanco" be "Mar Blanco"? (That seems to be what [Wikipedia says](https://es.wikipedia.org/wiki/Mar_Blanco).)

3. For South Africa, the Capital info has an an "and" instead of an "y" near the end.

(I've only very briefly glanced through your comprehensive list of issues, so sorry if there's something here that you've already brought up there.)

Other than that, everything looks perfect! :)

### Basic sanity-check

(just in case — it's easy to copy-paste into the wrong place.)

There are no changes to the existing data (there's only the addition of the Spanish translation).

<details>
```bash
# for master
csvtool cols 1- src/data.csv > old
# for the pr
csvtool cols 1,2-3,5-8,10-13,15-18,20-23,25-29,31-34 src/data.csv > new
diff old new
```
</details>

josealberto4444:
Minor issues fixed, thank you! =)

About the sanity check, there is indeed a difference. Tarawa was written as "Tarawa" even though it was only a word. I guess it was due to being edited by hand when changing from South Tarawa to Tarawa and the file was not reindexed.

josealberto4444:
I changed "Oficialmente conocido como Esuatini desde 2018" to "Oficialmente Esuatini desde 2018" because I think it's more correct and follows other strings as "Oficialmente Côte d'Ivoire". I hope you don't mind.

aplaice:
Thanks!

> About the sanity check, there is indeed a difference. Tarawa was written as "Tarawa" even though it was only a word. I guess it was due to being edited by hand when changing from South Tarawa to Tarawa and the file was not reindexed. That's the only difference. =D
> If you prefer it for a cleaner git log, I can "fix" the Tarawa issue in a single commit previous to the translation commits. I didn't do it before making the PR because it was such a minor issue.

Oh. The script didn't notice, since `csvtool` exported the data according to the "normal" (non-PHP) CSV quotation convention. In any case, I don't really see any benefit in separating it out — it's not a change in actual data, `git blame` etc. will be confused anyway and if anybody ever comes out with a nice `git` tool for tracking by cell (rather than by line/row) it should ignore quotation changes anyway.

> I changed "Oficialmente conocido como Esuatini desde 2018" to "Oficialmente Esuatini desde 2018" because I think it's more correct and follows other strings as "Oficialmente Côte d'Ivoire". I hope you don't mind.

It does seem better; thanks! :)

axelboc:
Amazing work, thanks a lot for your meticulous contribution!

Haven't had a chance to read [your comment](https://github.com/axelboc/anki-ultimate-geography/issues/161#issuecomment-583866082) and @aplaice's response to it in issue #161, so I'll leave the issue open for now and try to get to it asap.

Issue #164, opened 2019-11-06
Country info often excessively informative in Country -> Map cards


aplaice:
For some of the Country -> Map cards, the "Country info" (present on the front) can provide the entire answer for you. For example, Scandinavia:

![screenshot_2019-11-06_19:12:23](https://user-images.githubusercontent.com/16461345/68325894-a6336000-00ca-11ea-9841-f32bb97290f9.png)

Some other particularly egregious culprits are Melanesia and Athos, and slightly less egregious ones include the Indonesian islands, the constituent countries of the UK, Macau or Hong Kong.

![screenshot_2019-11-06_19:33:40](https://user-images.githubusercontent.com/16461345/68326782-62d9f100-00cc-11ea-92c3-4abaedcaa681.png)

## Some more examples, with country info, for reference

Country | Country info
----------|----------------
Melanesia | Subregion of Oceania, which includes the four countries of Vanuatu, the Solomon Islands, Fiji, and Papua New Guinea.
Polynesia | Subregion of Oceania comprising thousands of small islands in the central and southern Pacific Ocean.
Micronesia | Subregion of Oceania comprising thousands of small islands in the western Pacific Ocean.
Java | Island of Indonesia.
Mount Athos | Autonomous orthodox monastic state in northeastern Greece, also known as the Holy Mountain.
England  | Constituent country of the United Kingdom.
Macau | Special Administrative Region of China.

## Some possible approaches

1. **Remove the Country info from the front template** for the Country -> Map cards, and only display it on the back. Possible downsides:

   - For some cards, the Country info might be useful to identify the country (as opposed to pinpoint its location...). The main case I can think of is eSwatini (with info "Also known as Swaziland."), and perhaps North Macedonia, Myanmar and the DRC, where people might not have heard of the new name.

     - Counterpoint: by the time the Country -> Map card comes up, the learner will have been exposed to the "Country info" in the Map -> Country card (as well as, probably, in the Country <-> Capital and Country <-> Flag ones), so will hopefully already be aware of it.

   - When revealing the answer (displaying the "bottom" side) having the Country info appear in the upper half of the card, at the same time as the map appears in the lower half, might be disorientating to the learner.

2. **Place the "Country info" within a [hint field](https://apps.ankiweb.net/docs/manual.html#hint-fields)** for the Country -> Map front template. For cases like the Indonesian islands or the constituent countries of the UK, the "Country info" is an ideal location hint! Possible downsides:

[...]

PR #202, opened 2020-02-10
Fix duplicated country names in Norwegian and French

aplaice:
([Reported here](https://github.com/axelboc/anki-ultimate-geography/issues/161#issuecomment-583842852).) Thanks @josealberto4444!

## Disambiguate the two entities named Saint-Martin in French

I'm using the parentheticals from Wikipedia ([Saint-Martin (royaume des Pays-Bas)](https://fr.wikipedia.org/wiki/Saint-Martin_(royaume_des_Pays-Bas) and [Saint-Martin (Antilles françaises)](https://fr.wikipedia.org/wiki/Saint-Martin_(Antilles_fran%C3%A7aises))), rather than just "(Pays-Bas)" and "(France)", since:

1. It's generally better to rely on Wikipedia.

2. Sint-Maarten technically isn't part of the Netherlands, only the Kingdom of the Netherlands.

3. In mainland France, there are [apparently several communes called "Saint-Martin"](https://fr.wikipedia.org/wiki/Saint-Martin#France_m%C3%A9tropolitaine), so just using "Saint-Martin (France)" could be confusing for a French person.

(I don't think that this should carry over to the Spanish translation — 1. and 3. don't hold for Spanish, while 2. is arguably debatable.)

(Merging this should wait until after the Spanish translation is merged — I'll update/fix this, then.)


axelboc:
Took me a while, but it finally ticked in my brain as to what this issue is about. Isn't distinguishing between two "countries" with the same name the role of the _Country info_ field? Putting this information in parens as well will add redundant information to a few templates.

Which template(s) is(are) affected by the missing _Country info_ field? Perhaps we should reconsider the solution we picked in #164?

aplaice:
You're right, what's the right solution is non-obvious. I'll open an issue discussing the possible alternatives, but in the meantime I've taken out the trivial fix in the Norwegian deck into #211.

PR #232, opened 2020-03-26
Give same prominence to South Africa's three capitals

### LINKED ISSUE #207: Give same prominence to South Africa's three capitals
@josealberto4444 in https://github.com/axelboc/anki-ultimate-geography/issues/161#issuecomment-583866082:

> 14. For South Africa, the English capital is put as Pretoria with some capital info explaining that there are three capitals with legislative, executive and judicial power. In Spanish capital, I put Pretoria for consistency and translated the capital info, but the Wikipedia infobox says: "Ciudad del Cabo (legislativa), Pretoria (ejecutiva), Bloemfontein (judicial)". Also, I think the three cities should be mentioned in the usual order of branches of government: legislative, executive and judicial.

@aplaice's reply:

> Ugh. The infobox in English Wikipedia also says "Pretoria (executive), Bloemfontein (judicial), Cape Town (legislative)" (as do the German, French and Norwegian ones), so the same solution should be used for all of them. (Encyclopaedia Britannica also gives the three cities as capitals...)
>
> The two possible approaches are:
> - (The current, de-facto approach.) Give the first city in the infobox as the "capital", and mention the remaining ones in the "Capital info".
> - List all the cities as capitals.
>
> I don't have any strong opinion on whether the current approach should be kept or not. This should probably go in its own issue...

@josealberto4444, I'm interested to know more about your last sentence:

> I think the three cities should be mentioned in the usual order of branches of government: legislative, executive and judicial.

Do you have a source that mentions this? If so, it might be worth reconsidering our current approach. Otherwise, I say we stay consistent with Wikipedia's Spanish info box and use Cape Town as the main Spanish capital.

aplaice:
Fix #207

I'm mostly keeping the "Capital info" the same, since the information is already there.  (The only change is the removal of the description of Pretoria as the "formal" capital in the German.)

Mostly as a curiosity, see [this dissenting view on French Wikipedia](https://fr.wikipedia.org/wiki/Afrique_du_Sud#G%C3%A9ographie_administrative), claiming that only Capetown is really a capital of South Africa...


axelboc:
Haha, this paragraph makes some bold assumptions! I've marked it as needing citation for good measure. 😄

Issue #229, opened 2020-03-25
Check/correct updated German translation


aplaice:
@Erim24 if you have time to look at these, at some point, please feel free to mercilessly tear them up — I made most of them with the thought that something is better than nothing.

I've listed the translations and summarised the reasons for the changes here, for ease of access, but it probably makes sense to continue the discussion in the relevant threads.

### Capital info mentioning The Hague (issue #218, pull request #228)

(Amsterdam is the official capital of the Netherlands (according to the constitution), but The Hague is the seat of the government, the legislature and the supreme court. (See [Capital of the Netherlands](https://en.wikipedia.org/wiki/Capital_of_the_Netherlands.))

For Tanzania, we do mention that:

> While Dodoma is the official capital, Dar es Salaam is the de facto seat of government.

)

Final English "Capital info":
While Amsterdam is the official capital, The Hague is the seat of the executive and legislative branches.
Merged German "Capital info":
Amsterdam ist die offizielle Hauptstadt, aber Den Haag ist der Parlaments- und Regierungssitz.

### Kaliningrad Oblast (issue #208)

Should the German also include a parenthesised explanation of what an oblast is in the country info? (Currently, it's "Oblast der Russischen Föderation.")

[Direct link to suggested update](https://github.com/axelboc/anki-ultimate-geography/issues/208#issuecomment-591128050).

Suggested updated English "Country info":
Oblast (region) of the Russian Federation.
Suggested updated German "Country info":
Oblast (Gebiet) der Russischen Föderation. (see [German Wikipedia](https://de.wikipedia.org/wiki/Oblast_Kaliningrad))

### Capital info for Canary Islands (issue #226)

(Make clear that Santa Cruz and Las Palmas are two different cities, since it otherwise sounds like it might be a single name...)

[Direct link to translation suggestion](https://github.com/axelboc/anki-ultimate-geography/issues/226#issuecomment-602224372).

Suggested English "Capital info":
The capital is shared between the two cities of Santa Cruz de Tenerife and Las Palmas
Suggested German "Capital info":
Die zwei städte Santa Cruz de Tenerife und Las Palmas de Gran Canaria sind zusammen Hauptstädte der (Autonomen Gemeinschaft der) Kanaren.

### Cetinje (issue #205)

According to the Montenegrin constitution, Cetinje is officially designated as the "Old Royal Capital" and is also the seat of the president. This information is now being added to the English "Capital info".

[...]

PR #238, opened 2020-03-29
Correct/add German translations

Erim24:
I added the changes discussed in issue #229


axelboc:
Thanks Erim 💯 

I'm going to keep this PR open for reference, but I think it's safer if I open one PR per issue (EU, Montenegro, and so on) with all the translations included in each of them so we don't forget any.

axelboc:
I believe that all the changes have been merged in the deck. 👍

PR #234, opened 2020-03-26
Change some strings in Spanish translation

josealberto4444:
This solves the conflicts with the Spanish translation regarding #213

PR #245, opened 2020-04-01
Add capital of the European Union

### LINKED ISSUE #213: Add or remove EU's capital
Currently, the EU has a capital in the French version of the deck, but none of the others.

<hr/>

Replying to https://github.com/axelboc/anki-ultimate-geography/issues/161#issuecomment-587128665:

> Here are my thoughts regarding the capital of the EU:
>
>   - First, it seems like a shame to have to add capital hints to distinguish Belgium from the European Union.

That's true, though adding a hint like "Sovereign state" wouldn't (IMO) really harm the value of the existing Capital → Country card for Belgium.\*

\* I'm not sure what would be the best hint for the EU.  "Supranational organisation" would uniquely identify the EU, so would be sub-optimal.  Our usual counterpart to "Sovereign state" — "Not a sovereign state" — is technically correct, but mildly misleading.  Perhaps "Not a state" would "leak" the least amount of information?

Additionally, given that (at least outside Belgium), one is far more likely to hear about Brussels as a shorthand for the EU institutions or the EU as a whole, in the same way that one uses "London" for the UK or "Paris" for France, than as a shorthand for Belgium, I don't think that it'd be somehow wrong for the capital of Belgium to have to be disambiguated. (Absolutely no offence meant to any Belgians!)

>   - Second, it's not a straightforward change:

>       - Although English Wikipedia mentions Brussels as the capital
>         in the info box, French Wikipedia doesn't and instead lists
>         the various institutions, so the French translation would
>         have to not have the capital.

If we're to consistently apply our guidelines, the French translation should indeed have the capital of the EU removed. However, it should be added to the English, Spanish and possibly Norwegian versions, but not the German one. It'd be a bit of a mess, but not more so than the capital of the British Indian Ocean territory...

>       - One of the institutions, the Parliament, is located in
>         Strasbourg, so should this city be considered a secondary
>         capital and mentioned in the *Capital info* field?

IMO if the EU's capital is added, this definitely should be mentioned.  The fact that there's no official capital and that Brussels is only the de facto capital (possibly with a listing of the institutions that are seated there the seat of the Commission and Council), like in the case of Nauru/Yaren, should probably also be noted.

> I feel like we keep trying to fit a square peg in a round hole. Maybe the current model of using the *Country - Capital* and *Capital - Country* templates for subdivisions, along with capital hints, has reach its limit and we should brainstorm an alternative? 

Interesting, perhaps you're right.

> I don't know, maybe we could add a new template, or decide that if a subdivision's capital conflicts with a sovereign country's capital, we simply don't add it.  After all, this is what we do with flags (e.g. we don't have a flag for French Guiana *because* it's the French flag). Anyway, just a thought for now. If you don't think I'm going crazy, we can discuss further this in a separate issue.

OTOH, I think that we don't have a flag for French Guiana because none of the subdivisions of France have a separate flag\*. If only a specific subdivision of a country had the same flag as the country itself, but the remaining subdivisions had different ones, it would IMO be worth including the flag even for the subdivision that had the same flag.

\* French Polynesia and New Caledonia are partial exceptions, but even in their case it seems that the situation is rather complicated and controversial: [French Polynesia](https://en.wikipedia.org/wiki/Flag_of_French_Polynesia), [New Caledonia](https://en.wikipedia.org/wiki/Flags_of_New_Caledonia).

Also, I'm not sure whether ignoring capital conflicts between a country and one of its subdivisions would help us much, since the "Capital hint" field would still need to be present for cases like Aruba/Sint Eustatius or Guyana/Cayman Islands.

<hr/>

Possible ways forward:

1. Keep the current situation (EU has a capital only in French).

    - Would probably need to add a capital hint.

2. Add for English/Spanish/Norwegian (?), remove for French; include "Capital info" mentioning Strasbourg; add a capital hint.

3. Remove altogether.

axelboc:
Fix #213

### Summary of changes

- Add Brussels as single capital in English, German and Spanish (it was already present in French)
- Add three capitals in Norwegian as per [Wikipedia infobox](https://no.wikipedia.org/wiki/Den_europeiske_union)
- Add capital info to all languages as per https://github.com/axelboc/anki-ultimate-geography/issues/213#issuecomment-604039084
- Add "Sovereign country" / "Not a sovereign country" hints to _Belgium_ and the _European Union_ to all languages except Norwegian, taking into account the new Spanish formulation (cf. #234).
- Fix typo in capital of Belgium in Norwegian: ~Brüssel~ => Brussel

### To do

- [ ] @josealberto4444, are you good with the Spanish version of the capital info suggested by @aplaice?
- [ ] @Vages, can you please confirm that _Brüssel_ was indeed a typo in Norwegian?


aplaice:
I'm pretty sure that Brüssel was a typo, since the spelling Brussel is used on all of [Wikipedia](https://no.wikipedia.org/wiki/Brussel), [The Norwegian Ministry of Foreign Affairs' brochure](https://www.regjeringen.no/no/dokumenter/statsnavn_b/id87840/#bel) and [The Great Norwegian Encyclopedia](https://snl.no/Belgia), which are the three sources listed in [TRANSLATION_SOURCES.md](https://github.com/axelboc/anki-ultimate-geography/blob/master/TRANSLATION_SOURCES.md).

I'd be happier though if the Spanish were double-checked.

Also, though the summary is technically wrong, since the Norwegian doesn't need a "Capital hint" and hasn't had one added! :D

josealberto4444:
Yes, I agree! Sorry, I skipped commenting on that issue.

aplaice:
No problem, and thanks for chiming in!

I hope that you're doing fine, despite the general situation and your defence (if it's not being postponed :/)!

josealberto4444:
Yes, it was postponed until we come back to a normal situation. =)

Issue #220, opened 2020-02-28
Answer giveaway "Flag - Country" cards


ukanuk:
Some flags include the name of the country/continent/etc written on it, for example with Nicaragua, El Salvador, and Paraguay. When reviewing on my smartphone in bed at night (bad habit, I know...), I can clearly read the country name on the front side of the card. I would prefer to recognize the flag by its colors, layout, and symbols, but whenever I get the "Flag - Country" card for Nicaragua, my eye is immediately drawn to the text, and I can read the answer with zero effort. 

See my proposed solution below. For affected country notes it modifies the Flag field with a second SVG image and some hidden html. To the "Flag - Country" card type, it adds one class to the existing image div and adds three styles . If someone has a cleaner solution, I'd love to hear it!

Proposed solution, using Nicaragua as the example:

**Create new SVG for problematic flag**
To obfuscate "Nicaragua" and "Central America", I used the Inkscape effect "Filters > Distort > Roughen" menu. My modified file is only 0.1KB larger according to Windows Explorer.
*Before:*
![image](https://user-images.githubusercontent.com/59303202/75547054-64f60000-5a6d-11ea-8537-d90f9e002cb1.png)

*After:*
![image](https://user-images.githubusercontent.com/59303202/75546827-d2edf780-5a6c-11ea-9cc3-1774c4259866.png)

**Modify Flag field in note:**
```
<div class="hidden-field-note">Do not edit this field directly. Instead, edit its HTML (hotkey is Ctrl+Shift+X).</div>
<img id="flag-regular" src="ug-flag-nicaragua.svg">
<img id="flag-blurred" src="ug-flag-nicaragua-blurred.svg">
```
![image](https://user-images.githubusercontent.com/59303202/75547027-527bc680-5a6d-11ea-87fc-e71da37d08ba.png)

**"Flag - Country" card type: Add "blurred" class to flag div on front template**
```
<div class="image blurred" style="">{{Flag}}</div>
```
**"Flag - Country" card type: Add more styling**
```
/* For "Flag-Country" cards on countries like Nicaragua where country name is on the flag */
.hidden-field-note                           { display: none; }
.image:not(.blurred) #flag-blurred { display: none; }
.image.blurred          #flag-regular  { display: none; }
```
![image](https://user-images.githubusercontent.com/59303202/75546349-aedde680-5a6b-11ea-9a00-5ecda9f74ffd.png)

axelboc:
[...]

PR #246, opened 2020-04-02
Prepare to handle obfuscated/blurred images as discussed in #220

ukanuk:


src/style.css
```diff
@@ -27,6 +27,12 @@
   margin-top: .75em
 }
 
+/* Some flags (e.g. Guam) contain identifying words which can give away the answer */
```
horwitz, line 30: `Guam` → `Guam's`
`which` → `that` (or `, which`, depending on what is meant)

ukanuk: Thank you, done.

src/style.css
```diff
@@ -27,6 +27,12 @@
   margin-top: .75em
 }
 
+/* Some flags (e.g. Guam's) contain identifying words that can give away the answer */
+/* Card backs always show the unmodified image(s) */
+/* If a blurred image(s) is available, card front shows that instead */
+.image--front > img[src*="-blur"] + img { display: none; }
+.image--back > img[src*="-blur"] { display: none; }
```
aplaice, line 34: I'm not sure about this, since this is the way @axelboc had originally suggested it, and it is nice and concise, but I think that it might be more readable to split each of these into three lines, especially since that's the convention we follow elsewhere in `style.css`. In any case, it's just a question of formatting.
```suggestion
.image--front > img[src*="-blur"] + img {
  display: none;
}
.image--back > img[src*="-blur"] {
  display: none;
}
```

ukanuk: My vote would be to style everything automatically. [This StackOverflow answer](https://stackoverflow.com/a/55443013/6047827) suggests using [Prettier](https://prettier.io/) with [Husky](https://github.com/typicode/husky) to "set up a pre-commit hook which will automatically clean your files using Prettier before committing any new code."

It might not be worth it to set up the hook for it to happen automatically, but a trivially easy option for now would be to manually paste code into https://prettier.io/playground, click *Show options* in the bottom left to change the parser to *css*, and paste in the code to be formatted. That changes the styling to match what you recommend.

If the project were to adopt this as a general policy, note it would make this change at the bottom of the `style.css` file. There are configuration options, but I'm not sure whether they are all available from the online playground. 

*Changing from*
```css
.nightMode .info, .nightMode .type {
  color: #ccc;
}
```

*Changing to*
```css
.nightMode .info,
.nightMode .type {
  color: #ccc;
}
```

axelboc: Yep, let's do that. I'm a big fan of Prettier! I'll open a new issue to set up Prettier. In the meantime, just use the playground as you suggested. Thanks!

aplaice: > If the project were to adopt this as a general policy, note it would make this change at the bottom of the style.css file.

That makes sense!

aplaice:
I'd prefer not to merge such a relatively significant change myself at this point, but I think that this is ready.

axelboc:
Thanks @aplaice, I don't mind doing the merges. It gives me a chance to classify the changes with labels and milestones, and to keep the graph clean so I know what to write in the release notes. If I ever stop merging PRs for whatever reason, don't hesitate to take over, though ^^

PR #249, opened 2020-04-02
Update ug-flag-el_salvador.png to be based on latest from Wikipedia

ukanuk:
Main change is green mountain instead of yellow. This was noticed while working on #247.


aplaice:
Thanks!

PR #258, opened 2020-04-06
Make maps of Balkans and Saint Helena consistent with main style

aplaice:
Saint Helena's map could do with a zoom box, but I don't think it's *essential* since there are no other islands in the immediate vicinity, and Saint Helena's exact shape isn't that important.

The Balkan Peninsula map uses a slightly different choice of definition of the Balkans than the existing map, but I don't think it's a worse definition (our current map uses the Danube-Sava-Kupa line, while my map uses the Soča–Vipava–Krka–Sava–Danube line (which was what was in Natural Earth Data)). According to Wikipedia, "The northern border of the peninsula is variously defined.". (The definition used for the map in the infobox is the Danube–Sava–Soča one which is different yet...) In any case the differences are quite slight.

(See #1.)

<details>
<summary>Comparisons of the old and new maps</summary>
 
### Balkans

before

after

### Saint Helena

before

after

</details>


axelboc:
Amazing! ✨

lbolan136:
Culturally Vojvodina and Slavonia should be considered part of the Balkan Penninsula.

PR #260, opened 2020-04-07
misc text cleanup (mostly spelling and commas)

horwitz:


src/media/ug-flag-nepal.svg
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" width="205" height="250" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="-17.582 -4.664 71.571 87.246"><title>Flag of Nepal</title><desc>Coding according to the official construction in &quot;Constitution of the Kingdom of Nepal, Article 5, Schedule 1&quot;, adopted in November 1990</desc><use stroke="#003893" stroke-width="5.165" xlink:href="#a"/><path id="a" fill="#DC143C" d="M-15 37.57h60L-15 0v80h60l-60-60z"/><g fill="#fff"><path d="M0 35.43c6.6 0 11.95-5.35 11.95-11.95a12.802 12.802 0 0 1-5.37 6.32l-1.02-.76 2-1.5-2.42-.62 1.28-2.17-2.5.36.37-2.48-2.17 1.3-.62-2.45-1.5 2-1.5-2-.63 2.44-2.16-1.28.37 2.5-2.5-.37 1.3 2.17-2.44.62 2 1.5-1 .77c-2.37-1.4-4.3-3.56-5.38-6.32 0 6.6 5.35 11.95 11.95 11.95zM-5.76 53.03l-5.36-.66 3.25 4.3-4.97 2.12 4.97 2.1-3.25 4.3 5.36-.66-.66 5.37 4.3-3.25L0 71.63l2.1-4.98 4.32 3.26-.66-5.36 5.36.67-3.25-4.3 4.97-2.1-4.97-2.12 3.25-4.3-5.36.65.66-5.36-4.3 3.25L0 45.95l-2.1 4.97-4.32-3.25.66 5.36z"/></g></svg>
```
horwitz, line 1: "Shedule" → "Schedule"

CONTRIBUTING.md
```diff
@@ -173,7 +173,7 @@ The following guidelines apply to map images:
 
 ### Translation sources
 
-If you are contributing a new language, please add any sources to [`TRANSLATION_SOURCES.md`](https://github.com/axelboc/anki-ultimate-geography/blob/master/TRANSLATION_SOURCES.md), also explaining possible stylistic choices.
+If you are contributing a new language, please add any sources to [`TRANSLATION_SOURCES.md`](https://github.com/axelboc/anki-ultimate-geography/blob/master/TRANSLATION_SOURCES.md), also explaining any stylistic choices.
```
horwitz, line 176: "Possible" can be vague

CONTRIBUTING.md
```diff
@@ -120,7 +120,7 @@ The following guidelines apply to flag images:
...
 
-If the name of a country appears clearly on a flag, a second version of that flag may be provided, with the name of the country blurred out. The name should be blurred using [Inkscape](https://inkscape.org/)'s Gaussian blur effect as explained in #247. The blurred flag must be named `ug-flag-<country_name>-blur.<svg|png>` and placed in the `media` folder. A second HTML element must then be added to the _Flag_ field _before_ the existing HTML element. This allows the blurred flag to appear on the front of the country's _Flag - Country_ card.
+If the name of a country appears clearly on a flag, a second version of that flag may also be provided, with the name of the country blurred out. The name should be blurred using [Inkscape](https://inkscape.org/)'s Gaussian blur effect as explained in #247. The blurred flag must be named `ug-flag-<country_name>-blur.<svg|png>` and placed in the `media` folder. A second HTML element must then be added to the _Flag_ field _before_ the existing HTML element. This allows the blurred flag to appear on the front of the country's _Flag - Country_ card.
```
horwitz, line 123: Add "also" before "be provided" to clarify that both versions of the flag should be included.

CONTRIBUTING.md
```diff
@@ -49,7 +49,7 @@ Anki requires each note to have a unique identifier. When you add a note to the
...
 
-Anki Deck Manager has a very specific way of wrapping fields with double quotes  in `data.csv` to escape special characters (cf. [#129](https://github.com/axelboc/anki-ultimate-geography/issues/129)). Whether you edit the file by hand or through an editor, chances are you won't end up with double quotes in the same places. If you commit the file as is, the diff will be cluttered with changes that have nothing to do with your edits. To avoid this, run `composer index` before commiting your changes. This command has the side effect of normalising the escaping of fields in the entire file.
+Anki Deck Manager has a very specific way of wrapping fields with double quotes in `data.csv` to escape special characters (cf. [#129](https://github.com/axelboc/anki-ultimate-geography/issues/129)). Whether you edit the file by hand or through an editor, chances are you won't end up with double quotes in the same places. If you commit the file as is, the diff will be cluttered with changes that have nothing to do with your edits. To avoid this, run `composer index` before committing your changes. This command has the side effect of normalising the escaping of fields in the entire file.
```
horwitz, line 52: "quotes  in" → "quotes in" (remove double space)

CONTRIBUTING.md
```diff
@@ -133,7 +133,7 @@ A note's _Flag similarity_ field contains a list of countries, each followed by
 Flag similarities are always **mutual**: if flag A is similar to flag B, then flag B is similar to flag A. To determine whether two flags are similar enough to warrant mutual _Flag similarity_ information, their differences must first be **identified** and **classified**. The following classification applies:
 
 - **Critical differences (C)**
-  - presence/absence of decoration - i.e. symbol, coat of arms, etc.
+  - presence/absence of decoration - e.g. symbol, coat of arms, etc.
```
axelboc, line 136: Hehe, this one's come up before -- `i.e.` is valid here as it explains what is meant by the term `decoration`. Very, very subtle difference, but either is fine by me. 😄

horwitz: Ah—especially with the "etc." at the end, I just read it as a list of examples (of `decoration`s)!

src/media/ug-flag-nepal.svg
aplaice: Nice catch!

I feel slightly uncomfortable having non-automated changes to the media files, when it's not necessary (e.g. blurring, resizing etc.). I've now fixed the typo upstream (on Wikimedia), but I don't think it's the only change that's been made since we imported the flag.

Also, newer versions of SVGO just strip out the description...

Though, on the whole, there's no harm in fixing this here.

CONTRIBUTING.md
aplaice: I have no strong opinion on `i.e.` vs. `e.g.` here, though if we switch to `e.g.`, I think we should remove the `etc.` as it becomes redundant.

horwitz: Good point—the `etc.` is now gone.

PR #261, opened 2020-04-07
"Flag" and "Map" placeholder hints

### LINKED ISSUE #152: Visual answer type hints

ukanuk:
Fix #152 

Someone with an iPhone ought to double-check dark mode still works properly with this, as I changed code from PR #173 to fix an AnkiDroid regression introduced with fb7f4e8.

In addition, note I used https://commons.wikimedia.org/wiki/File:World_location_map_(W3).svg to get a proper outline of a Winkel III projection as found in this deck's inset maps, but don't think it's needed in `sources.csv` as I remade the file by hand to optimize size (original outline was 700 nodes, I remade by hand with just 6 nodes).

Lastly, I just noticed in [CONTRIBUTING.md](https://github.com/axelboc/anki-ultimate-geography/blob/master/CONTRIBUTING.md) that maps get a width of 500px, unlike flags which get a height of 250px. I will rectify this in a bit.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -37,6 +37,18 @@
   display: none;
 }
 
+.image__placeholder {
+  color: #333;
+  height: auto; /* Required for Chromium */
+  max-width: 100%;
+}
+
+.image__placeholder path {
+  fill: none;
+  stroke: currentColor;
+  stroke-width: 1;
+}
+
 .image img {
   box-shadow: 0 1px 4px 1px rgba(0, 0, 0, .2);
 }
@@ -53,6 +65,11 @@ hr {
   margin: 1.5em 0;
 }
 
-.nightMode .info, .nightMode .type {
+.night_mode .info,
+.night_mode .type,
+.night_mode .image__placeholder,
+.nightMode .info,
+.nightMode .type,
+.nightMode .image__placeholder {
   color: #ccc;
 }
--- a/src/templates/Country - Flag.html
+++ b/src/templates/Country - Flag.html
@@ -5,7 +5,17 @@
   <hr>
 
   <div class="type">Flag</div>
-  <div class="value">?</div>
+  <div class="image">
+    <svg
+      class="image__placeholder"
+      xmlns="http://www.w3.org/2000/svg"
+      width="417"
+      height="250"
+	  viewBox="0 0 417 250"
+    >
+      <path d="m2 26s45-24 129-24c86 0 146 25 210 25 64 0 74-10 74-10v218s-25 13-83 13-146-25-206-25c-60 0-123 19-123 19z" />
+    </svg>
+  </div>
 {{/Flag}}
 
 --
--- a/src/templates/Country - Map.html
+++ b/src/templates/Country - Map.html
@@ -4,7 +4,17 @@
   <hr>
 
   <div class="type">Location</div>
-  <div class="value">?</div>
+  <div class="image">
+    <svg
+      class="image__placeholder"
+      xmlns="http://www.w3.org/2000/svg"
+      width="500"
+      height="308"
+	  viewBox="0 0 500 308"
+    >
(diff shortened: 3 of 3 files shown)


aplaice:
Unfortunately, `.night_mode` is not recognised in older versions of Anki (2.1.17) — which is a mild inconvenience for end users, as it's easy to upgrade (and besides the "dark mode" in that version of Anki is half-baked anyway) — or of AnkiMobile (2.0.48) — which is a greater issue since newer versions of AnkiMobile are not available if you have a sufficiently old iOS device (I don't really use my old iOS device, only for occasionally testing like here, but there may well be people who do). I don't have access to a newer iOS device / a newer version of AnkiMobile.

I'd suggest including both classes (`.night_mode ...` and `.nightMode ...`), in both places.

ukanuk:
@aplaice great point, it never occurred to me that old iOS devices might be locked to old versions of the app. 

These last two commits are potentially controversial, and I'm not especially pleased that `hr` alphabetizes to the very bottom and `.night_mode` css all appears in the middle. But now with a lot of extra CSS compared to a month ago and the fact that one has to scroll to see everything regardless, alphabetizing it all seemed reasonable to me.

aplaice:
Thanks for fixing the night mode issue! (The CSS's a bit verbose, now, but there's no going around it.)

> These last two commits are potentially controversial, and I'm not especially pleased that hr alphabetizes to the very bottom and .night_mode css all appears in the middle. But now with a lot of extra CSS compared to a month ago and the fact that one has to scroll to see everything regardless, alphabetizing it all seemed reasonable to me.

It might be worth splitting out into its own issue or PR.

src/style.css
```diff
@@ -49,10 +44,37 @@
...
-.nightMode .info, .nightMode .type {
+.night_mode .info,
+.nightMode .info {
   color: #ccc;
 }
+
+.night_mode .type,
+.nightMode .type {
+  color: #ccc;
+}
```
axelboc, line 60: - The two night-mode rulesets could be merged into one since they both set the same colour.
- Actually, the night mode ruleset for the SVG placeholders could also be merged with the other two thanks to a little trick:
  1. Replace `stroke: #333;` in ruleset `.image__inline-svg path {}` with `stroke: currentColor;`
  2. Set `color: #333;` to `.image__inline-svg`.
  3. You can now write `.nightMode .image__inline-svg { color: #ccc; }` and merge it with the others.  
- It doesn't matter in practice thanks to selector specificity, but the resulting ruleset should come after the `.type {}` ruleset, since `.nightMode .type {}` is more specific. It might be worth just putting the whole ruleset right down the bottom of the file for simplicity.

axelboc:
Just read your comment about alphabetising the selectors. I get it now, but can't say I'm a huge fan because ordering in CSS matters. Let me know if the changes I suggested make things more readable overall.

ukanuk:
I've reverted the alphabetization to follow the order of the HTML elements, and I've also moved the `night_mode` stuff to the end. Great idea on using `currentColor` to combine all the `night_mode` selectors into one, I didn't know about that trick!

> * How about `image__placeholder` (or `value__placeholder`) instead of `image__inline-svg` for a bit more semantic?

My idea was that it ought to be `.image svg` to match `.image img`, but I renamed it to  `.image__inline-svg` to avoid confusion about whether it applies to  elements referencing an SVG file. But `image__placeholder` works fine, too, I've switched it.

> Now that the CSS is growing a little, I wonder whether it would be less confusing to merge classes `value` and `image`.

I think this would work and am happy to make the change, but it involves all templates and makes it harder to review the flag/map placeholder changes. So would it be better to put in a separate PR? This other PR could also provide space to discuss CSS organization as @aplaice suggested. At the very least, I won't merge any commits concerning this until you're able to review my changes addressing the rest of your feedback.

PR #272, opened 2020-04-10
Refactor class names and tidy styles

### LINKED ISSUE #270: Refactor class names and styles
As discussed in #261, I'd like to do a bit of tidying in the styles by replacing class `image` to `value` (and `value--image`).

axelboc:
Fixes #270

- Replace class `image` with classes `value value--image`.
- Rename `image__front` and `image__back` to `value__front` and `value__back`.
- Rename `image__placeholder` to `placeholder` for simplicity.
- Review comments in CSS and add more.
- Reorder CSS properties.
- Use direct descendant selector `>` where possible.
- Fix indentation in HTML to use spaces.
- Make images properly responsive with `max-width: 100%` and `height: auto` - not sure how this was working before 🤷

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -36,6 +36,8 @@
 }
 
 .value > img {
+  max-width: 100%;
+  height: auto;
   box-shadow: 0 1px 4px 1px rgba(0, 0, 0, 0.2);
 }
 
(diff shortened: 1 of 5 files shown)


aplaice:
Unfortunately, despite some thinking I haven't noticed any ways of further simplifying the CSS.

Just in case, (if case both of us missed something), I've tested the changes, and seen no issues!

PR #274, opened 2020-04-10
Don't apply shadow to images with non-rectangular outlines

axelboc:
The technique we used to show blurred flags made me think of this.

Some images don't have a rectangular outline, so the box shadow looks weird (see below). I suffixed the filenames of those images with `-nobox` and used it in the CSS to filter them out before applying the box shadow.

The affected files are the maps of the continents and oceans, and the flag of Nepal.

### Before

image

image

### After

image


aplaice:
This is a brilliant idea! Everything looks perfect!

## `filter: drop-shadow()` (?) (probably not worth it)

What about trying to use `-webkit-filter: drop-shadow()` (and possibly similar properties for people reviewing online on non-Safari browsers — I haven't looked into compatibility, yet)? It has the advantage that it applies to the outline of the underlying image, not just the bounding box, taking into account transparency. Unfortunately:

1. Most browsers (web engines) [don't support the `spread-radius` parameter](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/drop-shadow#Parameters) for `drop-shadow`, so we couldn't just use it as a global drop-in, though we could, in principle use it as a fallback, say:

```css
.value > img[src*="-nobox"] { /* nobox obviously becomes a misnomer */
  -webkit-filter: drop-shadow(1px 2px 4px rgba(0, 0, 0, 0.2));
}
```

2. Since the Ocean maps don't have a transparent background, they would still be stuck with an ugly "unattached" shadow.

3. The newly "noboxed" images have very little white content near the edges, so adding a shadow to them doesn't help with the purpose of bringing out white elements.

On the whole, I don't think it's worth bothering with.

### Flags with black elements in night mode (?)

This is a side-topic, inspired by your comment in `style.css` — what about flags with black elements in night mode. (If you agree that this is worth investigating, I'll open a new issue.)

src/style.css
```diff
@@ -38,6 +38,13 @@
 .value > img {
```
aplaice, line 38: Perhaps, now `.value > img` and `.placeholder` could be consolidated, e.g.:
```css
.value > img,
.value > .placeholder {
  max-width: 100%;
  height: auto;
}

.placeholder > path {
    color: #333;
    [...]
}

[...]
.night_mode .placeholder > path /* for consistency with the changes*/
[...]
```

axelboc:
@aplaice yeah I don't think it's worth using `filter` either.

Hadn't thought of putting a white shadow in dark mode, though, that sounds great!

PR #275, opened 2020-04-10
Improve map of Southern Ocean (slightly)

axelboc:
- PNG instead of SVG
- Same size as other ocean maps
- No weird top border


aplaice:
This is obviously an improvement! 

One possible issue is one of consistency — now the Southern Ocean map has a transparent background, while the remaining Ocean maps don't. Since a transparent background is definitely preferable to white one, I wouldn't worry about this, though.

PR #276, opened 2020-04-10
Refactor img and .placeholder

aplaice:
Since `.value > img` and `.value > placeholder` have the same functionality it might make sense to combine their declaration blocks.

Please feel free to reject/tear this up completely. :)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -35,7 +35,8 @@
   margin-top: 0.75em;
 }
 
-.value > img {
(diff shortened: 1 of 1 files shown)


axelboc:
Hmm why not! I might just bring back the `.placeholder{ color: #333; }` ruleset, though, as I find it a bit clearer (considering the comment and the night mode selector).

Issue #210, opened 2020-02-17
Clarify overall translations guidelines


axelboc:
[TRANSLATION_SOURCES](https://github.com/axelboc/anki-ultimate-geography/blob/master/TRANSLATION_SOURCES.md) should clarify the overall guideline of following the main Wikipedia site for each language when their contents differ (notably the article title and the capital(s) listed in the infobox).

We should also mention that:

- This guideline doesn't apply to flags, since translating the _Flag_ field seems a lot of effort for just a few minor differences (notable state vs. civil flags)...
- If there hasn't been discussions about the differences on Wikipedia, then it's always good to start one.

axelboc:
I think #255 made this issue a little redundant and makes me wonder whether `TRANSLATION_SOURCES.md` shouldn't be merged into `CONTRIBUTING.md`. I'm going to try this out.

PR #279, opened 2020-04-11
misc style updates

horwitz:


axelboc:
Ha, I was about to make similar changes as part of #210! 😄

Issue #221, opened 2020-03-17
Missing British Overseas Territories


ohare93:
Came across [Akrotiri and Dhekelia](https://en.wikipedia.org/wiki/Akrotiri_and_Dhekelia) while looking something up. Any reason this territory of Britain is not included? It has a capital "Episkopi".

Then looking at the wiki page for [British Overseas Territories](https://en.wikipedia.org/wiki/British_Overseas_Territories) I saw that we also do not have:

- [Tristan da Cunha](https://en.wikipedia.org/wiki/Tristan_da_Cunha) (Capital: "Edinburgh of the Seven Seas" :sunglasses:)
- [South Georgia and the South Sandwich Islands](https://en.wikipedia.org/wiki/South_Georgia_and_the_South_Sandwich_Islands) (Capital: King Edward Point)

Every other British Overseas Territory seems to be accounted for :+1: 

###### Tbh I want Tristan da Cunha in the deck just because "Edinburgh of the Seven Seas" is fucking badass :laughing:

aplaice:
I think that in general we have far too many tiny, underpopulated overseas territories, and perhaps we're overdue for a cull (see #137).

> Then looking at the wiki page for British Overseas Territories I saw that we also do not have:

What about [Ascension Island](https://en.wikipedia.org/wiki/Ascension_Island)? (population 806)

Also, what about the remaining French collectivities...

* [Wallis and Futuna](https://en.wikipedia.org/wiki/Wallis_and_Futuna) (population 11,558)
* [Saint Pierre and Miquelon](https://en.wikipedia.org/wiki/Saint_Pierre_and_Miquelon) (population 6,008)

...and the various (uninhabited) Antarctic territories...

For comparison, the populations of Akrotiri and Dhekelia, Tristan da Cunha and South Georgia and the South Sandwich Islands, are respectively:

* Akrotiri and Dhekelia: 15,700 (half of which is military personnel and their families)
* Tristan da Cunha: 246
* South Georgia and the South Sandwich Islands: 0 permanent, 99 non-permanent...

I'd vote for *possibly* including Akrotiri and Dhekelia, as having the largest population and some political significance/interest (e.g. the UK was offering to give up part of the base, as a sweetener for the Annan plan for the reunification of Cyprus in 2004), and ignoring the rest (sorry Edinburgh of the Seven Seas!), though that's just my arbitrary opinion.

ohare93:
[...]

PR #290, opened 2020-04-18
Improve night mode glow/shadow of flags

aplaice:
Implement @ukanuk's suggestions, resolving the problem with the glow's asymmetry, and its unnecessary visibility on a grey background.

I'm making the PR to get one step closer to fixing this, but @ukanuk, @axelboc and anyone else please feel free to tear it apart.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -90,7 +90,7 @@
  */
 .nightMode .value > img:not([src*="-nobox"]),
 .night_mode .value > img:not([src*="-nobox"]) {
-  box-shadow: 0 1px 4px 1px rgba(202, 202, 202, 0.2);
+  box-shadow: 0 0 4px 1px rgba(54, 54, 54, 0.9);
 }
 
 hr {


aplaice:
I feel rather bad that I hadn't thought of these changes before @ukanuk suggested them, and that they weren't included in the released version of the deck. (The asymmetry was/is a matter of taste that simply didn't bother me (though I now agree that unlike for a shadow, it doesn't really make sense), but I had been concerned by the unnecessary glow-on-grey...)

ukanuk:
I feel it's mostly my fault, as I was only subscribed to notifications for releases, and didn't happen to check the project for the few days right before the release, during which there was a flurry of changes including this one. I didn't notice the conversation until getting notified of the release, and reading the release notes, and clicking the linked issue to learn more (and updating to the release version to get all the changes and see them for myself).


Now I'm subscribed to be notified of all conversations, not just releases.

My guess is a lot more people are subscribed to notifications for just releases, and thus feedback will be much higher than normal immediately after a release. Maybe this feedback could be more quickly and cleanly integrated in the future by first releasing revisions as a beta, with a release note that feedback is especially welcomed for a two week period after which the full release will come.

Either that, or write a Github Action which can more easily allow smaller releases more often.

Just some thoughts. I think the current release process is fine, too. This is such an amazing collaborative deck already, and any ongoing work is just more icing on the top! My mind has been blown the last few months how CrowdAnki and Anki Deck Manager have enabled this significant community of anki editors to share their deck improvements. Who would have thought when starting this project that it would already been on a third major revision and have multiple releases each with a book chapter of improvements! So cool 😎

axelboc:
A pre-release sounds like an excellent idea for next time, you're totally right.

FYI, I'll try to release v4.4 in a month. What I have in mind for v4.4, apart from the usual fixes, is to figure out issue #137 (and #221) and perhaps integrate some more map improvements.

---

Are you happy for this PR to be merged by the way? Or would you like to further discuss the idea of limiting the use of the glow to only the flags that need it?

aplaice:
> Are you happy for this PR to be merged by the way? Or would you like to further discuss the idea of limiting the use of the glow to only the flags that need it?

Yes, definitely — you and @ukanuk convinced me that limiting it would not be useful and possibly harmful.

<hr/>

A pre-release is a great idea!

(Also, thanks @ukanuk for the words of support!)

ukanuk:
This PR additionally fixes #290

PR #292, opened 2020-04-19
Fix mistyped country names in French

### LINKED ISSUE #291: FR version: A few wrong names
I don't know how to propose a change to data.csv but:

Taiwan should be Taïwan
Péru should be Pérou
Quatar should be Qatar
Vénézuela should be either Venezuela or Vénézuéla

axelboc:
Fix #291


aplaice:
FWIW I can't find any uncaught occurrences of the country names and agree with your (implicit) decision not to mention the accented version of Venezuela in the country info, as it'd look silly.

axelboc:
Haha yeah, me too!! 😄

PR #304, opened 2020-04-30
Add tags for East and West Africa

### LINKED ISSUE #298: Add subregion tags for Africa
As discussed in #244 (https://github.com/axelboc/anki-ultimate-geography/issues/244#issuecomment-619439698), I think adding `UG::West_Africa` and `UG::East_Africa` makes the most sense.

aplaice:
Fix #298.

I've split the pull-request in two, for easier checking and because I'm not sure about the decision regarding East Africa.

### West Africa

The lists of countries tagged as belonging to West Africa:

Benin
Burkina Faso
Cape Verde
Ghana
Guinea
Guinea-Bissau
Ivory Coast
Liberia
Mali
Mauritania
Niger
Nigeria
Saint Helena
Senegal
Sierra Leone
The Gambia
Togo

Here, the [Wikipedia article](https://en.wikipedia.org/wiki/West_Africa) and the [Wikipedia category](https://en.wikipedia.org/wiki/Category:West_African_countries) are fully in agreement. (They also agree with the [UN geoscheme](https://unstats.un.org/unsd/methodology/m49/#geo-regions).)

### East Africa

Burundi
Comoros
Djibouti
Eritrea
Ethiopia
Kenya
Madagascar
Malawi
Mauritius
Mayotte
Mozambique
Réunion
Rwanda
Seychelles
Somalia
Somaliland
South Sudan
Tanzania
Uganda
Zambia
Zimbabwe

Here the [Wikipedia article](https://en.wikipedia.org/wiki/East_Africa) and the [Wikipedia category](https://en.wikipedia.org/wiki/Category:East_African_countries) are not in agreement — the former contains Sudan, while the latter doesn't.

I'm a bit disappointed about the article since it outright lies:

> In the United Nations Statistics Division scheme of geographic regions, 20 territories make up Eastern Africa:[1]
> [...]
>    Sudan and South Sudan – collectively part of the Nile Valley. Situated in the northeastern portion of the continent,[8] the Sudans are often included in Northern Africa.[9] Also members of the Common Market for Eastern and Southern Africa (COMESA) free trade area.

The UN Statistics division definitely does not place Sudan in Eastern Africa.

(When initially comparing the different regions of Africa, I had been misled by the relative confidence of this article (compared to, say, those for Southern or Central Africa, which are far more open about acknowledging the wildly differing definitions, rather than very openly favourising one of the definitions), and hadn't dug deeper.)

The article on Sudan itself also doesn't say that it's part of East(ern) Africa.

Hence, I've decided to follow the category rather than the article (excluding Sudan), but I'm not sure about this.

On the whole, I still think that East and West Africa make a good set of tags, though I'm less confident.


axelboc:
I agree with your assessment, the article is quite incoherent. I'm for following the UN geoscheme in this case and not include Sudan.

PR #309, opened 2020-05-16
Add more consistent maps of Sumatra and Java

### LINKED ISSUE #269: Improve maps of Java and Sumatra
> Identified in https://github.com/axelboc/anki-ultimate-geography/issues/1#issuecomment-610031655

If islands of Indonesia are to remain in the deck (cf. #137 and https://github.com/axelboc/anki-ultimate-geography/issues/221#issuecomment-607404670), bring maps of Java and Sumatra in line with locator map styles.

ug-map-java
ug-map-sumatra

aplaice:
Fix #269.

If the maps are OK, I'll upload to Wikimedia and update `sources.csv`.

<details>

### Java

before

after

### Sumatra

before

after

</details>

<hr/>

These unfortunately weren't done by the desired (almost) fully automated method, so adding more island maps won't scale too easily.

(The automated method doesn't work out-of-the-box due to a mismatch in the 50m-scale Natural Earth data for coastlines and for special features, such as labelled islands (this means that the borders of the islands don't *quite* line up with the coastlines...). There is no mismatch in the more detailed 10m data, but using that directly results in excessively large file-sizes for upload to Wikimedia. Reducing the detail in both sets of data in the same way (to avoid introducing the mismatch again) is tricky.)

As I had mentioned in one of the relevant threads (I don't remember which one...), I didn't just recolour the SVG map from which `ug-map-indonesia.png` was created, as I think that Indonesia is distorted there (it uses the W3 projection centred at 11°30' East, which doesn't work well for countries at "large" longitudes).


axelboc:
Looks great! Thanks for that. 🎉

aplaice:
> If the maps are OK, I'll upload to Wikimedia and update sources.csv.

Now, done!

In case we decide to include more physical islands, I've also uploaded maps for [Borneo](https://commons.wikimedia.org/wiki/File:Borneo_in_its_region.svg) and [Sulawesi](https://commons.wikimedia.org/wiki/File:Sulawesi_in_its_region.svg), as the marginal effort was negligible since they all used the same "frame".

(Bali will unfortunately require more effort since it would benefit from a zoom-box.)

PR #325, opened 2020-06-14
Update ocean maps to be similar to the continent ones

### LINKED ISSUE #264: Improve maps of oceans
> Identified in https://github.com/axelboc/anki-ultimate-geography/issues/1#issuecomment-610031655

Should ocean maps be made more consistent with locator maps? If they are fine as is, then I'll refocus this issue on the map of the Southern Ocean, which is inconsistent with the other ocean maps.

ug-map-atlantic_ocean
ug-map-southern_ocean

aplaice:
Fix #264.

It took longer than expected, due to various issues ("densification" of the long, straight borders between Oceans and anti-aliasing in SVGs (!)).

<details>
<summary>
Maps before and after
</summary>

### Arctic Ocean

before

after

### Atlantic Ocean

before

after

### Indian Ocean

before

after

### Pacific Ocean

before

after

### Southern Ocean

before

after

</details>

### Borders of the oceans

My main doubts are whether to follow the limits currently used in our maps (from the CIA World Factbook), where all the seas are considered part of the ocean to which they're adjacent (e.g. the Mediterranean is part of the Atlantic etc.).

According to the IHO's 1953 [Limits of Oceans and Seas](https://epic.awi.de/id/eprint/29772/1/IHO1953a.pdf) (which is the latest published (non-draft) version of the document and what we've *mostly* been following for the seas):

> The limits, given of the Arctic (17), Atlantic (23) and (32), Pacific (57) and (61), and Indian (45) Oceans exclude the seas lying within each of them

so according to it, all these the seas should be excluded from the relevant oceans.

However, that document does not include the Southern Ocean *at all*, so unless we were (crazily) to start omitting the Southern Ocean (which Wikipedia and most other sources do include) we couldn't follow it fully, anyway.

The [draft 2002 version of the IHO's Limits of Oceans and Seas document](https://web.archive.org/web/20190306055017/http://www.iho.int/mtg_docs/com_wg/S-23WG/S-23WG_Misc/Draft_2002/Draft_2002.htm) does include the Southern Ocean, but AFAICT it's also far less categorical about the limits of the oceans (which (if any) of the seas should be considered part of which ocean), so it's not too much help either. (Particularly see the introduction and the section on South East Asia.)

[Borders of the Oceans](https://en.wikipedia.org/wiki/Borders_of_the_oceans) has a relatively good overview of the topic (unlike the main Wikipedia articles (see below)).

FWIW Wikipedia's main articles on the oceans use the CIA World Factbook definitions, while claiming that they're following the IHO definitions (e.g. see [Atlantic Ocean](https://en.wikipedia.org/wiki/Atlantic_Ocean) (they claim the displayed CIA World Factbook map follows the IHO 2002 definition, (they don't provide a reference to the document, but if it's the above IHO draft, then the map and the document don't agree; as noted the document is not entirely clear on the borders of the oceans, but if anything, according to it, the Norwegian Sea should be considered a part of the Arctic Ocean, not the Atlantic, as the CIA map shows.)) or the [Arctic Ocean](https://en.wikipedia.org/wiki/Arctic_Ocean) (the article claims that the shown borders are "as delineated" by the IHO, without specifying which document they're referring to))... :(

Overall, I'd suggest sticking with the limits from the currently used CIA World Factbook maps. However, if you (or anybody else) have an alternative opinion, then adjusting the suggested maps would be straightforward, now that the generation script works.

<hr/>

### Central longitude

For the oceans other than the Pacific I've chosen a central longitude of 11.5 ° E, to avoid splitting Russia's Chukchi peninsula.

### Possible improvements

One improvement would be, for the Arctic and Southern Oceans, to include a view centred around the poles of the respective Oceans, as a sort of "zoom-box". However, I think that, contrary to my concerns, these two maps look OK-ish even without this.


axelboc:
I think they're fantastic! ✨

Happy to stick with including the seas until someone provides a source and reasoning for the contrary. Would you mind just posting the script (or a link to a Gist or whatever) somewhere for safe keeping, when you're ready, perhaps in a comment on the issue?

Let me know when it's ready to merge 😉

axelboc:
Oh just noticed that Greenland is cropped on the Pacific Ocean map -- any chance you could adjust the centring a little or was there a reason for the current centring?

aplaice:
> Would you mind just posting the script (or a link to a Gist or whatever) somewhere for safe keeping, when you're ready, perhaps in a comment on the issue?

The not-yet updated code dump is [here](https://github.com/aplaice/svg_sea_maps) (with fake history, since I didn't want all my messy commits to go public — it's enough that the horrible code is public...) It's an unholy mess of bash, python and make. For the purposes of generating the ocean maps it's a complete overkill, but for the sea maps — its original purpose — it made (some) sense, since I needed a way of combining various shapefiles (with various degrees of processing, since some of the tools used didn't deal well with some edge cases) in various combinations, with caching. 

I'll update it asap...

(Aside: I still need to upload the SVG maps corresponding to the sea maps, to Wikimedia. I'll probably have to ask again on Wikimedia on how best to proceed, since my previous query half a year ago was ignored and I don't want to upload ~ a hundred of 5 MB SVG files without any consultation.)

> Oh just noticed that Greenland is cropped on the Pacific Ocean map -- any chance you could adjust the centring a little or was there a reason for the current centring?

I used that centering since it's what was used for Pacific region "X in its region" country maps that I was emulating for the sea maps. I'll try adjusting it, but I'm not sure whether there is a good centering that doesn't crop anything, since the easternmost point of Greenland is east of the westernmost point of Iceland.

axelboc:
> I used that centering since it's what was used for Pacific region "X in its region" country maps that I was emulating for the sea maps. I'll try adjusting it, but I'm not sure whether there is a good centering that doesn't crop anything, since the easternmost point of Greenland is east of the westernmost point of Iceland.

Right, fair point. I think your initial centering was indeed the best.

> (Aside: I still need to upload the SVG maps corresponding to the sea maps, to Wikimedia. I'll probably have to ask again on Wikimedia on how best to proceed, since my previous query half a year ago was ignored and I don't want to upload ~ a hundred of 5 MB SVG files without any consultation.)

Arf, that's annoying. I hope they get back to you this time.

ohare93:
Looks great mate, nice effort 👍

aplaice:
I think that it's ready now, unless you notice anything wrong!

PR #333, opened 2020-06-20
Adjust colours of Greenland map

aplaice:
Stop highlighting Denmark.

For future reference colours adjusted with the barbaric `sed -e 's/fill="#E0E0E0"/fill="#FDFBE5"/' -e 's/fill="#FDFCEA"/fill="#FDFBE5"/'`.


axelboc:
Fantastic, thanks for that!

PR #334, opened 2020-06-20
Describe Bern as the de facto capital of Switzerland

### LINKED ISSUE #322: Capital of Switzerland
While the seat of the federal government is located in Bern, Switzerland intentionally does not have a capital, to avoid centralization. Currently, Bern is listed as the capital without further comments.

https://en.wikipedia.org/wiki/Switzerland

aplaice:
The descriptions are based on those for Nauru.  The Norwegian one was absent there, and hence wasn't added here.  (I'll open a separate issue for both.)

For ease of reference, the new `Capital info`s:

> Switzerland has no official capital; Bern is the de facto capital.

> Die Schweiz hat keine offizielle Hauptstadt; Bern ist der Regierungssitz.

Note that the German `Capital info` for Nauru (and hence now for Switzerland) refers to the "de facto" capital as the seat of government.  In these cases (and most others), the two are the same, but they're obviously not *quite* the same, in principle.  For both [Nauru](https://de.wikipedia.org/wiki/Nauru) and [Switzerland](https://de.wikipedia.org/wiki/Schweiz), the capital is described as *de facto* in the infobox (and it's stated that there's no *de jure* capital); for Switzerland, Bern is also explicitly listed as the seat of government (Regierungssitz).  I think that it definitely makes sense to use the same structure for both, and most probably makes sense to keep the current one.

> Suiza no tiene capital oficial; Berna es la capital de facto.

> La Suisse n'a pas de capitale officielle ; Berne est la capitale de fait.

Fix #322.


axelboc:
You're on a roll, @aplaice 🚀😄 Trying to get as much stuff into v4 as you can? Happy to postpone it by a week if you prefer! ^^

aplaice:
> Trying to get as much stuff into v4 as you can? Happy to postpone it by a week if you prefer!

Yeah, though I'm not sure if there's much sense in postponing, as the low-hanging fruit is gone. :) Adding the newly-included political entities can definitely wait, especially since I doubt that I could get nice maps done for the ones that lack them, so not all of them would be added, anyway. (Also, finalising the translations might take time, for some of them.)

Vages:
> Sveits har ingen offisiell hovedstad. Bern er regjeringssete.

Would be the Norwegian translation. 🇳🇴

axelboc:
Alright, I'm on it, thanks!

aplaice:
Thanks @Vages for the translation and @axelboc for finishing this off!

Issue #307, opened 2020-05-12
Upgrade process when data is removed (v4 upgrade)


axelboc:
Discussion started in #306. The problem with re-importing a deck with fewer notes/cards is summarised in https://github.com/axelboc/anki-ultimate-geography/issues/306#issuecomment-623368100.

The goal is to find an upgrade process for users who may want to keep the notes/cards that are removed from the deck. The process would have to impact "normal" users as little as possible.

One approach, based on releasing a major version, was explained in https://github.com/axelboc/anki-ultimate-geography/issues/306#issuecomment-623398459. Obviously, this impacts "normal" users significantly, so it's far from ideal.

aplaice:
I agree with @ohare93 and @ukanuk that in principle, a better deck management system would *probably* be ideal. (Sorry, I still haven't had time to look at Brain Brew!)

<hr/>

That said, I don't think that using the currently available tools are insufficient for a "good enough" outcome.

To summarise, we have three categories of notes:

**I.**  Those which will be kept, with all their current fields. (The bulk)

**II.** Those which will be kept, but with the contents of some of their fields removed (the "Capital" and "Flag" fields), hence causing some of their cards to become empty and probably, eventually deleted. The cards:

* **II.a.** The cards that will remain (Map→Country and Country→Map (extended deck))

* **II.b.** The cards that will be functionally removed from AUG (Country⇔Capital, Country⇔Flag).

**III.** Those which will be completely removed.

To slightly modify/re-phrase @axelboc's [categorisation of scenarios](https://github.com/axelboc/anki-ultimate-geography/issues/306#issuecomment-623398459):

#### 1: New user or upgrading user who doesn't mind losing their existing progress and any now-removed cards. Wants to follow AUG in choice of cards (they don't want the now-removed cards).

They want to only have the cards from **[I.]** and **[II.a.]**.

This group must be able to just import the new deck (possibly deleting their old copy first), without having to fiddle with anything.

#### 2: Upgrading user who'd prefer not to lose progress if possible, but wants to follow AUG in choice of cards.

They want to only have **[I.]** and **[II.a.]** and lose as little progress as possible.

#### 3: Upgrading user who'd prefer not to lose progress if possible, and doesn't want to lose any now-removed cards. Wants to stay "inter-operable" with AUG.

They want to have all of **[I.]**, **[II.a.]**, **[II.b.]**, **[III.]**.

(I'm not sure that group **3** is *much* smaller than **2**, if at all.)

[...]

PR #337, opened 2020-06-21
UG Legacy - temporary deck for v4 migration

axelboc:
I've modified UG in this branch in order to build the temporary deck for the v4 migration, as discussed in #307 (option D). I won't merge this PR, obviously, but at least it's here, in GitHub's history, in case somebody ever needs it.

For reference, I have:

1. Replaced `src/data.csv` with `removed/data.csv`, and `src/media` with `removed/media`.
2. Changed the GUIDs of the notes that have been partially removed from UG (the ones starting at line 19 in `data.csv`). I did this by removing their GUIDs then running Anki DM's `index` command.
3. Created copies of the standard and extended decks' configurations with Anki DM. The copies have their own deck GUIDs, but I left their config and model GUIDs unchanged. The copies are called `Ultimate Geography Legacy` and `Ultimate Geography Legacy [Extended]`.
4. Updated the deck description to link to the future v4.0 release.

---

I've tested the process on my own (standard English) deck and it went very smoothly! I imported first the temp deck then the v4 deck (built from `master`), and then used Anki's `Empty Cards...` feature, and everything's as expected. Most importantly, **I did not lose any progress on the main deck**! I also checked for unused media files and only those that have been renamed in v4 showed up.

I then deleted the legacy deck, checked for unused media files, which listed the correct files. I don't have an extra note type, and everything still looks great in the main deck! 💪  

FWIW, with option D in #307, I don't think it matters whether the temp deck is imported before or after the v4 deck. I'll still tell people to import it first, but I won't insist on the import order.

I'll move on to updating the README, the deck description and so on, and then I'll write the pre-release notes. I'll leave this PR open until the official v4 release. In the mean time, feel free to have a go at importing the temp deck and let me know if you notice any issue.


ohare93:
Perhaps this is relevant, just 20mins ago I checked in a feature to CrowdAnki in my [Personal Fields + Import Config Window PR](https://github.com/Stvad/CrowdAnki/pull/72) to allow users to select whether they wish to import the Note Models, Notes, and Media, each independently :tada: I have tested it and it works like a charm. I believe the Note Model option will be the most useful :grin: 

image

Though it is not pulled in yet, Stvad has been active reviewing my PR this last week, so it's only a matter of time! May be worth waiting for :crossed_fingers:

axelboc:
Happy to hear that! Not sure I understand how this relates to the v4 upgrade, though. Can you please elaborate? I'm almost done writing the pre-release notes, so I wouldn't be very keen in changing the current plan at this point. 🤔

ohare93:
Oh you are literally releasing about now? Then don't wait up, go ahead :+1: 

I merely meant that users would have more control over what to import, and this is a rather delicate import :thinking: but if all is well with the import, and everyone else has the same experience, then there's no need for that anyways :sweat_smile:

src/desc.html
```diff
@@ -1,16 +1,5 @@
+<b>Ultimate Geography v4 (LEGACY)</b>
 
-<a href="https://github.com/axelboc/anki-ultimate-geography/"><b>FULL DESCRIPTION</b></a> | <a href="https://github.com/axelboc/anki-ultimate-geography/releases"><b>RELEASE NOTES</b></a> | <a href="https://github.com/axelboc/anki-ultimate-geography/blob/master/CONTRIBUTING.md"><b>CONTRIBUTING</b></a>
+This deck holds content that was removed as part of Ultimate Geography's v4 release. If you're not interested in this content, you may delete this deck.
```
aplaice, line 3: Should there perhaps be slightly more detail here? 

In principle, if somebody imports the legacy deck, then it means that they'll have seen the release notes, and hopefully read them, but if they did things in a hurry, they might possibly get confused (e.g. why are there "new" cards?). Obviously, if they get confused they should click on the link, but I know that I myself can behave irrationally when panicking...

Perhaps:

```suggestion
This deck holds content that was removed from the main deck as part of Ultimate Geography's v4 release. It will no longer be updated by the project. If you're not interested in it, you may delete this deck.

Some of your progress for the cards in the removed content has had to be reset, which is why some cards are marked as "new". (No progress will have been lost in the main deck.)
```

(I'd add "has had to be reset for technical reasons", but that sounds awfully weaselly...)

Obviously, this is just a suggestion!

axelboc: Sounds much clearer, thanks 👍

Issue #143, opened 2019-09-15
Investigate better internationalisation format/system/workflow


axelboc:
The current CSV format is not adequate for managing translations:

- The diffs are unreadable.
- Changes to translations conflict with content edits.
- The growing number of columns makes `data.csv` difficult to work with.

CSV was good to start with, but I think we need to investigate more advanced internationalisation formats/systems/workflows if we want to encourage more contribution.

Vages:
Could not agree more with your description of the problem.

Anki Deck Manager (anki-dm) seems to have one source of truth: `data.csv`. It is possible (and beneficial) to solve the problem without modifying anki-dm. I may be missing something, but I think the simplest way to do this on our own is to split translations over several files and merging them with a script. If you agree with the suggested solution, only two questions remain: (1) Which format should be used to store the split data? (2) How will the merge script work (algorithm/language)? 

Here are my suggestions for answers to these two questions:
(1) CSV should be understandable for less techy users and is the result file format. So I think it's the best file format contender. Only other format that's as widely read by different libraries and users is JSON, but I don't think nested data structures (such as JSON) are well suited for tabular data. So CSV wins that head-to-head fight.
(2) Merging the files should be a simple join operation, as seen in every Databases 101 subject. Each row already has a unique ID, which could be used to refer to each item across files. Merging the files should be solvable in most high-level languages. PHP is the strongest contender, as it is already used in the project. If we want to avoid PHP (which I don't think we should want to, but anyway), my suggestion is to use Javascript (Node) or Python, as they are likely to already be installed on the user's computer.

Do people agree?

Vages:
... (2) cont'd: A bash script could also be a contender: https://stackoverflow.com/questions/6301059/how-to-join-2-csv-files-with-a-shell-script However, I don't know how compatible this is with Windows.

Vages:
And as another aside: If we go for a PHP/CSV combination, our solution is more likely to be transferred into the anki-dm core than if we choose some other combination.

Vages:
This Stack Overflow answer seems to be of some help (to PHP noobs like myself): https://stackoverflow.com/questions/15490465/how-to-combine-rows-with-the-same-id-from-a-csv-with-php

ohare93:
[...]

PR #341, opened 2020-06-22
Create dummy data.csv with only English and empty "xx" columns

aplaice:
I've created the dummy `src/data.csv` as promised!


aplaice:
Also (to avoid polluting the main issue), you were right about the GUIDs! (I had always incorrectly assumed without checking that CrowdAnki's and Anki-DM's GUIDs, for the same notes, were the same...)

ohare93:
The GUIDs are indeed different in the deck compared to the data csv, as I commented [upon doing my first move to Brain Brew test branch](https://github.com/axelboc/anki-ultimate-geography/issues/143#issuecomment-614053311). This is because Anki-DM takes the values in the guid column and runs them through a guid generator, in this case double guid-ifying them! :smile: In practice you could simple put in the country name into there, and it should all work fine :sweat_smile: 

I have wrote a little Python script to copy the *actual* GUIDs over, which I used last time, so I will just copy over all your changes again when I make the proper BB deck :+1:

PR #356, opened 2020-08-15
Add Jeju

### LINKED ISSUE #315: Add political entities with maps as per new inclusion rules
The following political entities are to be added to the deck, **with maps only** (and country info as needed), as per the new [inclusion rules](https://github.com/axelboc/anki-ultimate-geography/pull/312) documented in [CONTRIBUTING.md](https://github.com/axelboc/anki-ultimate-geography/blob/master/CONTRIBUTING.md#political-geography) and [political-entities.xlsx](https://github.com/axelboc/anki-ultimate-geography/blob/master/political-entities.xlsx):

- [x] [Wallis and Futuna](https://en.wikipedia.org/wiki/Wallis_and_Futuna)
- [x] [Akrotiri and Dhekelia](https://en.wikipedia.org/wiki/Akrotiri_and_Dhekelia)
- [x] [Svalbard](https://en.wikipedia.org/wiki/Svalbard) (i.e. the unincorporated area)
- [x] [Autonomous Region of Bougainville](https://en.wikipedia.org/wiki/Autonomous_Region_of_Bougainville)
- [ ] [Jeju Province](https://en.wikipedia.org/wiki/Jeju_Province)
- [x] [Hawaii](https://en.wikipedia.org/wiki/Hawaii) (i.e. the state)
- [x] [Alaska](https://en.wikipedia.org/wiki/Alaska)

For consistency, the locator maps should be centred on the territories and not highlight their parent countries.

aplaice:
Fix #315.

~If the map is suitable, I'll upstream it and add it to `sources.csv`.~ (now done)


axelboc:
Hey, thanks for that!!

I like it, but do you think the island should perhaps be a bit more centred within the zoom box (i.e. a bit further from the corner)? I don't think having the whole of South Korea in the box is necessary, so it could probably be a bit more square (i.e. less tall) and moved towards the South a bit. It might make it a bit more consistent with the other locator maps, what do you reckon?

aplaice:
Unfortunately, I can't move the box further south, as the [map](https://commons.wikimedia.org/wiki/File:South_Korea_location_map.svg) that I'm adapting for the zoombox doesn't go any further south. I can't find any other map of the region, of comparable detail and even the most fine-grained Natural Earth data isn't as precise, so I couldn't recreate it, easily. (Jeju is pretty tiny...)

I could cut off as much of Korea, as necessary, though:

ug-map-jeju

The square zoombox does look nicer. OTOH, I feel as if the zoombox somehow loses context, now. (My extremely weird thought every time I look at the inset zoombox, is "what is this small island off the coast of Ireland (and why is France so close)?")

axelboc:
>I can't move the box further south, as the map that I'm adapting for the zoombox doesn't go any further south. 

Ha, makes sense.

> "what is this small island off the coast of Ireland (and why is France so close)?"

So true! 😂 I think you're right, since the box can't be moved further South, your initial version looks clearer. 👍

aplaice:
> and therefore from the bottom of the locator map

Perhaps, the zoombox should be in the upper right corner? Jeju will still be in the bottom of the zoombox, but not of the map as a whole:

ug-map-jeju

(Initially, I had placed the zoombox in the lower right, since it wasn't "blocking" anything other than mostly empty sea.)

axelboc:
Ah, good idea, have you tried centred as well? So there isn't too much empty ocean below the box.

aplaice:
> Ah, good idea, have you tried centred as well? So there isn't too much empty ocean below the box.

Like this:

ug-map-jeju

?

You're right, it might be the best.

axelboc:
Perfect! Thanks for the tweak ;)

Issue #248, opened 2020-04-02
Set up Prettier


axelboc:
Suggested in #246 by @ukanuk to format `src/style.css` automatically.

To discuss/investigate:

- I don't know whether Prettier comes as a PHP package, so we may have to use NPM and add a `package.json` file. This means adding `npm install` as an additional set-up step for contributors.
- Prettier might be useful to format more files than just `style.css`. It doesn't seem to support CSV, but it does support HTML, JSON, Markdown, etc.
- A pre-commit hook might add unnecessary overhead, as most changes occur in `data.csv`. A CI job could be sufficient, IMO.

axelboc:
Might be worth waiting to see what's happening with Brain Brew (cf. #143).

ukanuk:
There may be a way to get this to work with Github Actions such as the third-party [Prettier Action](https://github.com/marketplace/actions/prettier-action) on the Marketplace. I got the action working on a personal repository, but I am concerned whether it will work for pull requests raised from forks. I realized this potential limitation while reading the [autopep8](https://github.com/marketplace/actions/autopep8) action docs.

Autopep8 suggests using a "ChatOps" slash command style as shown in [slash-command-dispatch](https://github.com/peter-evans/slash-command-dispatch), which I'm thinking could potentially be triggered by the fork author. But I have not tested this. Another thought I had was maybe the GitHub action could add a comment to the pull request with the issues as in [Spectral comment](https://github.com/marketplace/actions/spectral-comment).

Maybe there's even a way to have it suggest the code changes [like aplaice did for me](https://github.com/axelboc/anki-ultimate-geography/pull/246#pullrequestreview-386237539), and I was able to implement his changes just by clicking "Commit suggestion".

axelboc:
I think it's still worth waiting for Brain Brew at this point. Perfect formatting is not crucial, so I think we can wait to see how our set-up changes before doing anything.

PR #364, opened 2020-10-04
Move To Brain Brew

ohare93:
Rather than discuss this further in #143, I am making this PR now so we can go over the still to do tasks before a Pull could happen. This also allows us to see the extent of the changes, in Git's change view.

So, what is still missing? :grin:


ohare93:
I am not an expert of Git Actions, so I will not mess with the previous system that ran the `composer index` check.

The build recipe can be verified with `brain-brew [recipe] --verify` and I am open to adding a full --dry-run option. Perhaps just building the deck parts that are saved and checking for diffs would also help.

aplaice:
AFAIR, the ultimate goal is to move to a single `data.yaml`, with one line per translated field. 

However, I think that it might be worth switching to `brain-brew`'s CSV system (as here/in your branch) before then, since it's still (on net) an improvement compared to the single CSV (`src/data.csv`). Also, the overall tooling won't change significantly when we switch formats "within" `brain-brew`, so it's not like we'll be forcing our contributors to learn a new system, unnecessarily.

IMO the most important thing missing is having the CSS and templates as separate files.

I have doubts whether per-field CSVs (`capital.csv ` etc.) are better than per-language ones (`french.csv` etc.) — I can see some advantages of either — but as long as we're planning to eventually switch to the YAML it's probably not worth obsessing over which to choose.

<hr/>

Also, congratulations on (soon) becoming a father!

axelboc:
I also don't mind if we keep the data in CSV files for now. It actually makes the transition smoother I think. 

> IMO the most important thing missing is having the CSS and templates as separate files.

Yeah, I agree -- merging without this would be a step back, I think. That being said, it's not a huge deal either, so if it postpones the move to BB too much, I wouldn't mind merging without it.

aplaice:
This looks fantastic!

Unfortunately, I won't have time for more than the cursory look I've had (or to test the conversion pathways/recipes) in the nearest future, but I think that it'd be great to get this merged soon!

Three things that might be valuable (but I don't think are essential before merging (not that it's up to me :smiley:)):

1. Remove the duplication of data, (particularly `src/data/split`, `src/data/combined_data.csv`, `src/deck_parts/notes/english.yaml`, leaving only the first), to avoid confusing contributors and to avoid stale data.

2. (Wishlist) have a recipe to "unsplit" the CSVs — in general the huge single CSV is unwieldy, but occasionally it's quite valuable. (i.e. the reverse of `legacy_csv_to_split_csvs.yaml`).

3. Update `CONTRIBUTING.md`.

ohare93:
> This looks fantastic!

Thanks! I just have a quick change to make so the deck description can live as it's own Html file too, and a bugfix about the importing media going in the right subfolder instead of the top level media folder. Then I'll release the latest version of Brain Brew, and point UG towards that :+1: (currently still running off my latest local branch).


> Unfortunately, I won't have time for more than the cursory look I've had (or to test the conversion pathways/recipes) in the nearest future, but I think that it'd be great to get this merged soon!

No problem mate, but I hope so too!

> Three things that might be valuable (but I don't think are essential before merging (not that it's up to me smiley)):
> 
>     1. Remove the duplication of data, (particularly `src/data/split`, `src/data/combined_data.csv`, `src/deck_parts/notes/english.yaml`, leaving only the first), to avoid confusing contributors and to avoid stale data.
> 
>     2. (Wishlist) have a recipe to "unsplit" the CSVs — in general the huge single CSV is unwieldy, but occasionally it's quite valuable. (i.e. the reverse of `legacy_csv_to_split_csvs.yaml`).
> 
>     3. Update `CONTRIBUTING.md`.

1. The combined_data.csv and the transformer are only there for one last transform of the current state of the data before the proper merge. Though I can leave the recipe there afterwards. Notes can be deleted too, no worries.

2. Sure, I'll throw that recipe together. Should be no problem.

3. Can do :+1: if you have anything specific in mind, let me know :+1:

aplaice:
Great! :)

> Can do +1 if you have anything specific in mind, let me know +1

I was thinking about something equivalent to the "install" and "running" steps from your [Brain Brew Starter Readme](https://github.com/ohare93/brain-brew-starter/blob/master/README.md), to replace the `anki-dm`-specific information.  (Hooray, we'll be mostly able to get rid of the "quote normalisation" issues!)  In the long-run we'll probably end up with some brain-brew-specific "best practices" and idiosyncrasies, to add there, but probably not just now. :)

ohare93:
Alright, all apart from the update to Contributing is now done :+1: 

I have released Brain Brew 0.2.5, which brings the changes I mentioned above:
- Deck Description is it's own Html file now
- Existing media will stay in any arbitrary folder inside the "src/media" folder when being overwritten on source_from_anki
- Notes can be given an override for the NoteModel (used in the Extended Deck to have the same notes, but a different Note Model, without redoing the work)

I have also added the `legacy_csv_from_split_csvs` recipe, which took literally 1 min to make :grin: as well as disabled most of the unused `save_to_file` options for now. The old combined csv remains, but that is because I believe it will be used before the proper merge.

The Extended Deck works, each language is working too. All the html and css files live on their own (though the downside is these lone html and css files cannot be written to in a `source_from_anki` build just yet, but eventually).

This is a great time for me to update my own Brain Brew documentation, as so much has changed since I last touched it, it is very out of date. I will in tandem update the Contributing here, but would like some sanity checking that everything seems to work first. No rush, it can wait until the new year. I hope Brain Brew will be used in UG by the end of January! :grin: :clap: 

Lemme know if you guys spot anything, whenever you have time. Thanks again for your hard work :+1:

aplaice:
Wow, that was amazingly fast!  I'll try to have a detailed look ASAP and definitely before the New Year (I hope...)!

ohare93:
> Wow, that was amazingly fast! 

Only a few months! :laughing: I finally had a good amount of free time today, for once in the last few months :wink:

ohare93:
Oh, I should of course do a pull of the latest UG again. And I am sure an update to the Workflow files would be appreciated too.

ohare93:
Pulled the latest, and updated the workflow to do a full build and check for errors :+1: 

Just have to make changes to the Contributing. At a glance I can see I should:
- Update `Setup`
- Update `Build and Import`
- Remove `Indexing`
- Remove `Quote normalisation`
- Update `Release process`
- Make Brain Brew specific notes
    - The files being separated
    - Guids are created automatically on build
    - Link to documentation on recipe files
    - Running `source_to_anki` or `source_from_anki`
    - Steps to add a new language


See anything else I should change? :grin: I'll get on this in a week or so, moving house today/tomorrow :see_no_evil:

ohare93:
Contributing updated :+1: latest pulled again too. Merry Christmas, all! :christmas_tree: :tada:

aplaice:
A late Merry Christmas to you and everybody, too!

<hr/>

This looks great! I also love the new, detailed section in `Contributing.md`.

Some non-exhaustive (I haven't had time to stress test too much) feedback:

### GUID generation

> #### Adding a new country
>
> Add the country to `main.csv`, and then any other derivative files that are needed.
>
> Do not create a guid for the country, that will be autogenerated on the
> first run of the `source_to_anki.yaml` recipe.

This doesn't seem to work (a guid is generated, and inserted into the `CrowdAnki` JSONs, but not into `main.csv`) .  Surprisingly, running *only* `brain_brew source_from_anki.yaml` does generate a guid.  Slightly weirdly, if you first run `brain_brew source_to_anki.yaml` and then `brain_brew source_from_anki.yaml` you get *two* lines corresponding to the newly created country. 

### Carriage returns

When using the `source_from_anki.yaml` (and `legacy_csv_to_split_csvs.yaml`) recipes, I still get Windows newlines (inserted carriage returns), in all the split CSVs, despite being on Linux.  

(This is obviously inconvenient, since it would mess up all diffs, unless we ran something like `git diff --name-only | while read i; do sed 's/\r//' -i "$i"; done` every time.)

I think I have a fix, though.

### `legacy_csv_from_split_csvs` 

Not that it really matters, but it doesn't seem to work.  Since this is a) a luxury, b) a luxury that I had asked for, and c) it would do me good to get re-acquainted with the recipes' YAML format, I'll look into this myself (at some point).

<hr/>

(Note: I'm obviously complaining about the few things that don't work :) )

.github/workflows/integrity-check.yml
```diff
@@ -9,27 +9,17 @@ jobs:
...
-
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
     - name: Install dependencies
-      run: composer install --prefer-dist --no-progress --no-suggest
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
```
axelboc, line 19: Any reason why the set up is different locally and in the CI? (i.e. why the CI needs a `requirements.txt` file and doesn't use `pipenv`?)

CONTRIBUTING.md
```diff
@@ -35,27 +35,120 @@ Seen something outdated or plain wrong? Spotted a typo somewhere? Think somethin
...
+
+#### Media
+
+Media files are stored in the `src/deck_parts/media` folder, under arbitrary subfolders.
+
+
+### Source to Anki
+
+From here, building the deck is as simple as running `brain_brew recipes/source_to_anki.yaml`.
+[Brain Brew][Brain Brew] builds the deck into the `brain_brew_build` folder,
```
axelboc, line 91: Could the folder just be called `build`?

CONTRIBUTING.md
```diff
@@ -35,27 +35,120 @@ Seen something outdated or plain wrong? Spotted a typo somewhere? Think somethin
...
+On first run it will generate all the missing files and folders, with some warnings in the output.
+This is normal.
+
+
+### Anki to Source
+
+One can also make changes in Anki, and pull the changes back into the csv files. Simply:
+
+1. Make your edit(s) in Anki
+1. Export the deck using CrowdAnki into the `brain_brew_build/Ultimate Geography` folder
```
axelboc, line 103: I wonder if it would be better to have a dedicated folder. It feels weird having a "build" folder used as source. Doesn't really fit with how Brain Brew works, but it would make more sense to me if the path to the CrowdAnki JSON was provided via the command line - e.g. `brain_brew recipes/source_from_anki --source path/to/json`. Anyway, doesn't really matter for now, just a thought.

.github/workflows/integrity-check.yml
```diff
@@ -9,27 +9,17 @@ jobs:
...
 
-    - name: Index data.csv
-      run: composer index
+    - name: Validate
+      run: |
+        brain_brew recipes/source_to_anki.yaml || (echo "::error file=source_to_anki,line=0,col=0::Build failed" && exit 1)
```
axelboc, line 23: Not sure I understand the error handling part of this command or why it's needed. Doesn't `brain_brew` itself log the error and exit with `1`?

axelboc (commented):
Happy New Year!! 🎉

This looks awesome! I just have a couple of suggestions on my end, notably with regard to the folder structure.

I reckon we shouldn't version-control `combined_data.csv` and the legacy recipes. I feel like we'd be better off storing the legacy recipes as gists and linking them from a discussion thread.

The GUID generation issue reported by @aplaice is a little problematic, but to avoid delaying the merge, we could just say in `CONTRIBUTING.md` that Brain Brew doesn't yet support adding new entries.

Since the YAML files are quite long, it would be nice to have some sort of automatic formatting in here as per #248. Is there a tool similar to Prettier (a code formatter that supports Markdown, YAML, etc.) but written in Python that we could install along with Brain Brew? It would be a shame to have to add a `package.json` file and `node_modules` folder just for Prettier.

aplaice:
Happy New Year!

> I reckon we shouldn't version-control combined_data.csv and the legacy recipes. I feel like we'd be better off storing the legacy recipes as gists and linking them from a discussion thread.

I fully agree that the `combined_data.csv` should stop being version controlled (if I recall correctly @ohare93 was planning to do it as a last step before merging, but it could just as well be done now), but the "legacy recipes" (i.e. being able to combine and split the CSVs on-demand) are IMO useful even going forward. The single CSV is often unwieldy, but it can also be convenient to have everything in a single file, so if we can (in effect) have both options (split and combined) at low cost, it'd be ideal!

> The GUID generation issue reported by @aplaice is a little problematic, but to avoid delaying the merge, we could just say in CONTRIBUTING.md that Brain Brew doesn't yet support adding new entries.

:+1: (Or use the workaround of counterintuitively running `brain_brew source_from_anki.yaml` first).

> It would be a shame to have to add a package.json file and node_modules folder just for Prettier.

We'll have just removed the dependency on PHP, so we can afford to be extravagant! :p Seriously, I have no idea about a good python yaml formatter. Googling brings up [yamlfmt](https://github.com/mmlb/yamlfmt), but it doesn't appear to be too popular (which given my lack of experience in the area would be an approximate proxy for potential longevity and quality).

src/deck_parts/media/flags/ug-flag-anguilla.svg
```diff
@@ -0,0 +1 @@
...27 1.416-21.471 2.334-19.177 4.091-45.373-16.193-61.488-9.778-5.776-4.15-6.965-10.606-6.732-19.307-12.554 10.598-6.272 21.575-9.129 26.805-7.397 12.748-16.83 25.497-11.827 42.248 3.798 12.716 6.629 10.197 5.447 20.166-1.198 6.441-.562 23.856 3.925 18.758z"/><path fill="#f90" d="M931.763 290.302c-13.703-1.828-34.694 5.451-43.059 5.967-2.778-8.347-1.51-8.97-12.893-14.605 4.032 14.055 4.922 14.776-.013 19.504-20.39 4.868-41.432-20.773-41.04-34.226-.009-13.678 5.401-8.097 6.946-27.707-10.543 3.357-8.485 15.424-14.502 20.874-6.66 1.02-11.413-8.837-20.74-10.422-3.058 6.111 9.613 12.426 13.201 16.311 13.592 13.53 8.935 45.397 23.029 55.227-.738 6.86-5.918 11.17-13.85 15.48 16.009 4.76 22.613-6.043 28.785-6.417 15.27-.56 31.594.542 44.043-12.205 9.45-9.676 5.724-10.682 15.285-14.875 6.401-2.355 21.705-11.882 14.808-12.906z"/><circle fill="#fff" cx="266.239" cy="122.104" r=".806" transform="matrix(3.50845 0 0 3.34488 6.33 -163.011)"/><circle fill="#fff" cx="266.239" cy="122.104" r=".806" transform="matrix(-1.639 -2.95744 3.10207 -1.56258 914.323 1224.328)"/><circle fill="#fff" cx="266.239" cy="122.104" r=".806" transform="matrix(-1.80698 2.86712 -3.00733 -1.72273 1749.003 -240.213)"/></svg>
```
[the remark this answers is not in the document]
aplaice: Done [here](https://github.com/ohare93/anki-ultimate-geography/pull/7). I don't feel too comfortable commenting on the rest, but I can at least do this. :D

.github/workflows/integrity-check.yml
aplaice: Perhaps to avoid installing pipenv in the CI, which adds overhead/increases build time, for little benefit, since compartmentalisation/not "polluting" the global namespace isn't valuable in the CI? (@ohare93 can confirm or deny!)

src/deck_parts/media/flags/ug-flag-anguilla.svg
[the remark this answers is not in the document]
ohare93: Did I accidentally include them back in in the merge? If so, my bad. Thanks @aplaice! :heart:

.github/workflows/integrity-check.yml
ohare93: Exactly as @aplaice says. The requirements file is just made from a freeze of the pipenv anyways :+1: It is also a handy file to have in the repo as Git picks it up and recognises dependencies from it, to show in the Using/Uses section.

.github/workflows/integrity-check.yml
ohare93: It does indeed return an error, but this was the only way I could get the Git CI to recognise it as a failure in another project :shrug: just copied it from there. If someone has a simpler solution I will happy steal! :smiling_imp:

ohare93:
Thanks for the feedback guys! :clap: 

> GUID generation
>
> This doesn't seem to work (a guid is generated, and inserted into the CrowdAnki JSONs, but not into main.csv) . Surprisingly, running only brain_brew source_from_anki.yaml does generate a guid. Slightly weirdly, if you first run brain_brew source_to_anki.yaml and then brain_brew source_from_anki.yaml you get two lines corresponding to the newly created country.

Ahh yes, I see I broke this when I split everything up the first time. Because the recipe no longer writes to the csv (as it was wasteful) it no longer updates the Guids. Hmm :thinking: I think I will create a "create_guids" task and just add that to the source_to_anki recipe. Or would you prefer another recipe specifically for this? Same thing, in the end.

It would be nicer to have a smarter component along the lines of "parse_file" with if_empty then create_guid and many options for which columns and what data blahblah but that's a bit complicated just to fix this little thing. I would like to have this functionality eventually though, at which point I will switch this over.

> Carriage returns
>
> I think I have a fix

Thanks for the fix! And the two other ones to this repo you also provided :handshake: :clap: 


>legacy_csv_from_split_csvs
>
>Not that it really matters, but it doesn't seem to work. Since this is a) a luxury, b) a luxury that I had asked for, and c) it would do me good to get re-acquainted with the recipes' YAML format, I'll look into this myself (at some point).

Hmm, it works for me. It may require you to do it on an empty legacy csv file :thinking: I will test this and get back to you with updated instructions.

> Since the YAML files are quite long, it would be nice to have some sort of automatic formatting in here as per #248. Is there a tool similar to Prettier (a code formatter that supports Markdown, YAML, etc.) but written in Python that we could install along with Brain Brew? It would be a shame to have to add a package.json file and node_modules folder just for Prettier.

I am also unsure of this, but we should definitely look into this to keep things nice and neat :smile: as @aplaice said, we can always just install Prettier! :laughing: no worries though

> I reckon we shouldn't version-control combined_data.csv and the legacy recipes. I feel like we'd be better off storing the legacy recipes as gists and linking them from a discussion thread.

Alright, good stuff. Here's what I'll do:
 - Remove the old legacy files
    - Turn the legacy csv into a Gist
    - Leave the legacy recipes
 - Change the folders as you requested 
    - remove the "split" and "deck_parts"
    - change "brain_brew_build" to "build" (was only named that to avoid collision with the old one :wink: )
    - Update all the recipes to point to the new locations of files
       - Make sure it all works again!

How does that sound? :+1:  

> Also, I hope that moving house went well! (I'm amazed that you manage to get so much done, while also being a father and moving houses, in Christmas season!)

Was a busy time. Boxes are still everywhere! :sweat_smile: Hopefully things speed up soon, though :wink: Thanks for the continued help :+1:

aplaice:
> Hmm thinking I think I will create a "create_guids" task and just add that to the source_to_anki recipe. Or would you prefer another recipe specifically for this? Same thing, in the end.

I think either would be fine. Perhaps having it in a single recipe/build step would be best for simplicity/convenience? (I've now spent an exorbitant amount of time thinking over this tiny detail and am still undecided...)

Pros for having a single build step:

1. Simple/convenient.
2. (In rare cases) avoids shooting oneself in the foot (if one builds a `CrowdAnki` json with a new note, imports the json into Anki, starts using it and only realises after several days (e.g. after a collaborator points it out) that the GUID was missing in the source CSV, one would need to extract the GUID manually from the JSON build (or even Anki, if one deleted the builds), for pasting into `main.csv`, to avoid losing one's Anki "progress").

Cons:

1. Mildly against "unix philosophy"/separation of concerns.
2. Increases the execution time, if one is actually interested in carrying out one of the steps (say because one only wants to add GUIDs, without building, or because one is sure that all fields already have GUIDs). OTOH `brain-brew recipes/source_to_anki.yaml` takes ~ 5 s, so it's not a monstrous slow-down!

(For comparison, `anki-dm`'s `composer build` does not add GUIDs, if necessary, but instead raises an error if they are missing and tells the user to run `composer index`.  On the whole, I think I'd prefer automatic addition of GUIDs — i.e. everything in a single recipe.)

<hr/>

>    Not that it really matters, but it doesn't seem to work. Since this is a) a luxury, b) a luxury that I had asked for, and c) it would do me good to get re-acquainted with the recipes' YAML format, I'll look into this myself (at some point).
> Hmm, it works for me. It may require you to do it on an empty legacy csv file thinking I will test this and get back to you with updated instructions.

I'll try again now.

<hr/>

> Alright, good stuff. Here's what I'll do:
>  - Remove the old legacy files
>     - Turn the legacy csv into a Gist
>     - Leave the legacy recipes
>  - Change the folders as you requested 
>     - remove the "split" and "deck_parts"
>     - change "brain_brew_build" to "build" (was only named that to avoid collision with the old one :wink: )
>     - Update all the recipes to point to the new locations of files
>        - Make sure it all works again!

That sounds perfect!

ohare93:
>     2. (In rare cases) avoids shooting oneself in the foot (if one builds a `CrowdAnki` json with a new note, imports the json into Anki, starts using it and only realises after several days (e.g. after a collaborator points it out) that the GUID was missing in the source CSV, one would need to extract the GUID manually from the JSON build (or even Anki, if one deleted the builds), for pasting into `main.csv`, to avoid losing one's Anki "progress").

Would never happen as, by my understanding, CrowdAnki will throw an error for an empty Guid. But I agree we should just generate the guid, as getting the vague error in CrowdAnki afterwards would be confusing.


> OTOH `brain-brew recipes/source_to_anki.yaml` takes ~ 5 s, so it's not a monstrous slow-down!

Yours takes 5 seconds? Mine is below 1 :laughing: :muscle: :smirk:

axelboc:
> On the whole, I think I'd prefer automatic addition of GUIDs — i.e. everything in a single recipe.

Yeah me too. What matters, I think, is that it's done by a dedicated task, so I'm all for having a new `create_guids` task for this, @ohare93 💯

aplaice:
> Would never happen as, by my understanding, CrowdAnki will throw an error for an empty Guid.

The issue would be that a GUID _is_ generated, by `recipes/source_to_anki.yaml` and entered in the `CrowdAnki` json, just not fed back into the CSV.  Hence, CrowdAnki would happily import the deck and the new note, into Anki, but with a GUID whose value was not registered in `main.csv`, and if a GUID were later generated in the CSV, it wouldn't match the one in Anki. (Obviously,  deleting the new "bad-GUID" notes in Anki, or manually copying the GUID(s) from the JSON to the CSV, would normally not be an issue, but it might be confusing and/or inconvenient.)

> Yours takes 5 seconds? Mine is below 1

:laughing:

ohare93:
> The issue would be that a GUID _is_ generated, by `recipes/source_to_anki.yaml` and entered in the `CrowdAnki` json, just not fed back into the CSV. Hence, CrowdAnki would happily import the deck and the new note, into Anki, but with a GUID whose value was not registered in `main.csv`, and if a GUID were later generated in the CSV, it wouldn't match the one in Anki. (Obviously, deleting the new "bad-GUID" notes in Anki, or manually copying the GUID(s) from the JSON to the CSV, would normally not be an issue, but it might be confusing and/or inconvenient.)

Oh shit, you're right, the guid gets generated automatically regardless :thinking: hmmmmmmm. I suppose I should make the notes_from_csv task stop generating the guids automatically, and have it return an error. Alright, this is an easy fix :+1: Will get to it soon.

ohare93:
All changes made :+1: 

The files should be in the locations discussed :+1: 

The Guid generation works when running either to or from source :+1: (side effect: now all the child csvs are ordered alphabetically, and will continue to update themselves to this whenever you run one of the recipes. Can be changed :+1: )

The Legacy Csv is now in a Gist https://gist.github.com/ohare93/0496ec1624d72abeb2085cb640d5528a :+1: though I had some issue getting the Legacy recipes to work without the actual file. Long story short, the csv builder needs to know what the headers are before it can populate the file. I'll make a generic task to get around this later on, but not now, since this is not that high priority. Of you can to use the legacy file you just need to make `src/data/combined_data.csv` and add the current headers, then everything should work fine :+1: 

Remember to update your version of Brain Brew in your own repo of this, if you wanna try it out :+1: I hope we can move forward and get this rolling! :grin: new translations and extension repos await! :grin:

src/note_models/style.css
```diff
@@ -90,7 +90,7 @@
  */
 .nightMode .value > img:not([src*="-nobox"]),
 .night_mode .value > img:not([src*="-nobox"]) {
-  box-shadow: 0 0 4px 1px rgba(54, 54, 54, 0.9);
+  box-shadow: 0 1px 4px 1px rgba(202, 202, 202, 0.2);
```
axelboc, line 93: Was this change made on purpose?

CONTRIBUTING.md
```diff
@@ -274,8 +367,8 @@ Content changes, such as adding a note, replacing an image, or translating the d
 ### Release process
 
 1. Bump the version in `desc.html` and commit the change.
-1. Run `composer index && composer build`.
-1. Add each folder in the `build` directory to a separate ZIP archive named as follows:
+1. Run `brain_brew recipes/source_to_anki.yaml`.
+1. Add each folder in the `brain_brew_build` directory to a separate ZIP archive named as follows:
```
axelboc, line 371: Just a remnant of the old name of the `build` folder.

src/data/country.csv
```diff
@@ -0,0 +1,312 @@
+country,country:de,country:es,country:fr,country:nb
```
axelboc, line 1: The column headers in the CSV are all in lower case, but the recipes refer to fields with their capitalised names. Doesn't matter, really, since it works fine as is, but just thought I'd point it out in case you hadn't done this on purpose.

axelboc (approved):
Fine by me! Ready to merge when @aplaice gives his green light 👍

src/note_models/style.css
ohare93: :thinking: nope, not by me. Might have happened during a merge, somehow. Will rollback if @aplaice confirms it is wrong :+1:

CONTRIBUTING.md
ohare93: Damn, I thought Pycharm found all the references. Fixed :+1:

src/data/country.csv
ohare93: I did not do it on purpose, but it is no matter as the csv generator lowers all filenames for comparison purposes. Feel free to lower them in the recipe too :+1:

axelboc: Meh doesn't matter for now. Maybe the generator should eventually respect the casing when writing the CSV, though?

src/note_models/style.css
aplaice: Yeah, this is wrong! (It's the old pre-65566644a61f08fffb81aa2854c7af1523ab3931 version.)

Good catch @axelboc!

aplaice:
This looks great! I can't wait for the new translations and for the extensions! :)

One more thing that we all somehow managed to miss is that the wrong templates were picked for the non-extended deck! We want the `Country - Capital`, `Capital - Country`, `Flag - Country`, and `Map - Country` notes rather than the `Country - Capital`, `Capital - Country`, `Country - Flag`, and `Country - Map` ones! (Unless we're celebrating switching to our brilliant new system, by increasing the difficulty of the deck, a notch! :) )

(I only noticed this when comparing our old `build` output with the new one (~ `diff <(jq --sort-keys < old_build/Ultimate\ Geography/Ultimate\ Geography.json) <(jq --sort-keys < build/Ultimate\ Geography/Ultimate\ Geography.json) ` (the output is a bit messy since only `dict`s are sorted, not lists)...)

ohare93:
> (Both fixed here.)

Lovely, thank you. That was my bad :sweat_smile: 

> Edit 2: I've also only now noticed that the deck names are now all "Ultimate Geography" (rather than "Ultimate Geography[xx]" or "Ultimate Geography [Extended]"). This is definitely not a blocker and I'm not actually certain which "style" I prefer.... (Using just "Ultimate Geography" for all the decks has the slight disadvantage that it might lead to some slight inconsistencies when re-exporting with CrowdAnki, since the newly generated folders/JSONs will always be called Ultimate Geography/deck.json rather than Ultimate Geography[xx]/... etc., but that's not really important.)

Ahh yes, as they all use the same Header. Hmm :thinking: I agree it would be good to have an override for the name here, but fyi the source_from_anki recipe currently _only_ looks at the deck `Ultimate Geography` so users of the translations must change the recipe to get the importing back into source. An improvement I'd like to do on this in the future is to have the deck name be Regex, and the builder offers all options that match in the chosen folder for selection mid run. But that is neither here nor there, this original deck name issue can be fixed easily in the future :+1:

aplaice:
> Ahh yes, as they all use the same Header. Hmm thinking I agree it would be good to have an override for the name here, but fyi the source_from_anki recipe currently only looks at the deck Ultimate Geography so users of the translations must change the recipe to get the importing back into source. An improvement I'd like to do on this in the future is to have the deck name be Regex, and the builder offers all options that match in the chosen folder for selection mid run. But that is neither here nor there, this original deck name issue can be fixed easily in the future +1

Yeah, given that the Anki → CSV contribution pathway is still relatively immature (we didn't have this available at all until now! :)) it's totally not an issue; I had simply noticed the (harmless) deck name change and commented!

ohare93:
@aplaice no problem mate, thank you for being so thorough :handshake: 

Alright, any other stragglers? Can we *push the button?!* :scream:

aplaice:
I think it's just the CSS and we can push the button (if GitHub allows us\*).

\* it's not happy that it has run only one test and it has grayed out the merge button...

ohare93:
CSS reverted :+1:

src/note_models/style.css
ohare93: Reverted :+1:

aplaice:
Great, thanks, hooray! I'm voting for merging! :)

aplaice (approved):

aplaice:
I'm still stumped by GitHub requiring two passing tests — after all, we've deliberately reduced the number of tests...

axelboc:
Sorry for the delay 😄

Amazing effort, @ohare93. 💯 Thanks a lot for getting this done!!! The future is bright. 😄

ohare93:
💯 🥳 ❗ 

>2 checks passed 

What magic did you do?! 😆

axelboc:
Haha no clue! 😄 You never know what's going to happen when you add/modify/remove a workflow in a PR...

PR #365, opened 2020-11-11
Add UK countries as dependent territories in inclusion rules

### LINKED ISSUE #361: Constituent countries of the UK not covered by inclusion rules
On Wikipedia, _England_, _Scotland_, _Wales_ and _Northern Ireland_ are listed neither as [sovereign states](https://en.wikipedia.org/wiki/List_of_sovereign_states) nor as [dependent territories](https://en.wikipedia.org/wiki/Dependent_territory#United_Kingdom). This is not consistent with the constituent countries of Denmark and the Netherlands, which are listed as dependent territories.

I suggest we mention them explicitly in the contributing guidelines, [under _Dependent Territories_](https://github.com/axelboc/anki-ultimate-geography/blob/master/CONTRIBUTING.md#dependent-territories), and add them to the _political-entities.xlsx_ spreadsheet. I've checked: they all pass the flag/capital threshold (by far 😄).

axelboc:
Fix #361


aplaice (approved):
This looks great!

<hr/>

<nitpicking_leading_to_nowhere>

I'm not quite sure whether the constituent countries of the UK strictly count as dependent territories. Unlike in the case of the constituent countries of the Kingdoms of the Netherlands or of Denmark, they're not partially independent states, that mostly just share foreign policy/defence. Scotland, Wales and N. Ireland are probably closer to autonomous regions, while England doesn't even have a parliament of its own.

However, given that people generally consider these subdivisions as "important" (more so than, say the Lands of Germany etc.), that the four territories have their own "national" sports teams and that'd it be a shame to remove them, I think that it makes sense to include them under the "dependent territories" umbrella. (Creating a separate category just for the four would feel a bit silly.) Hence, I think that the change is perfect! :) 

</nitpicking_leading_to_nowhere>

PR #370, opened 2020-12-21
Fix spelling of Nur-Sultan in Spanish

axelboc:
Follow up to #351: as you expected, @aplaice, Spanish Wikipedia now has [Nur-sultán](https://es.wikipedia.org/wiki/Nur-sult%C3%A1n) as the article title, but it includes a hyphen.


axelboc:
Ignoring the failed CI checks, which seem to be caused by a newer version of PHP on the CI. With Brain Brew coming, doesn't seem worth fixing.

PR #371, opened 2020-12-21
Rename Kiev -> Kyiv (Official Name)

nakutnyi:
Consider renaming Ukrainian capital Kyiv.
This complies with the guideline as Wikipedia page is already updated.
See https://en.wikipedia.org/wiki/KyivNotKiev for more info.


axelboc:
Thanks @nakutnyi!

aplaice:
Yes, definitely!

The change was made on [16 September](https://en.wikipedia.org/w/index.php?title=Kyiv&type=revision&diff=978662036&oldid=978647643) after a [protracted discussion](https://en.wikipedia.org/wiki/Talk:Kyiv/Archive_7#Requested_move_28_August_2020), I think that it's safe for us to follow WIkipedia — it seems unlikely that Wikipedia will revert the change.

nakutnyi:
@axelboc @aplaice thank you! You are awesome!

Issue #313, opened 2020-05-20
Dutch translation


hoi3:
Thanks for the amazing geography deck! 
I am considering making a Dutch translation of the deck. However, I am not at all familiar with GitHub, Anki Deck Manager and other aspects that are involved in the process. I've read the CONTRIBUTING.md but still I'm not sure how to proceed exactly. I have installed the CrowdAnki addon and Anki Deck Manager but I'm a bit lost from there on. I If you could explain a bit more detailed how I should approach this project, I think I'd be able to make my translation publicly available. I am planning to make Dutch translations anyway, so I think it would be worth it for me to put a little extra effort into this so that other people can use the translated deck as well.

axelboc:
That's very nice of you! 💯 

I'd love to tell you how to get started but I'm afraid adding a sixth translation may not be the best idea at this point in time... 😓

Firstly, we're currently in the process of removing a bunch of content from the deck - cf. #312. So it'd be a shame for you to translate something that'll end up being removed.

Secondly, because of the way the deck is currently set up, with Anki Deck Manager, all translations are managed within a single CSV file: `src/data.csv`. That's five languages cramped into a single CSV file... I won't get into the details, but in short, we've reached the limit of our set up... Just open it and you'll understand. 😄 

But hope is in sight! One of our contributors, @ohare93, is working on a new deck management system called _Brain Brew_, which should make it much easier to manage translations. If you'd like to know more, check out issue #143. I can't give you an estimate as to when we'll migrate to Brain Brew or when we'll be ready to integrate new translations -- all I can say is that it may very well take a while.

---

Now, it'd be a shame for you to translate the deck anyway, as you seem to want to do, and not be able to share it!

The way I see it, if you could get your translations into a CSV file similar to `data.csv` and upload it to this issue, then I'm sure we'd manage to integrate them into the deck once it's ready to receive more translations. Sure, some of the translations would become outdated, but I reckon it wouldn't be too much effort to go through and update them when the time comes.

How to get your translations into a CSV file is really up to you. To get started, you'd probably want to proceed as follows:

1. First wait for PR #312 to be merged, or better, if you're patient, for v4.0 of the deck to be released, which should happen within a month.
1. Make a copy of the deck's `src/data.csv` file.
1. Make some space in the file by deleting the columns that belong to other translations (`Country:de`, `Country:es`, etc.).
1. Add the appropriate columns for your translation: `Country:nl`, `"Country info:nl", etc.
[...]

PR #378, opened 2021-01-20
Added Dutch translation

WouterVdS:
#313 Dutch Translation


recipes/legacy_csv_to_split_csvs.yaml
```diff
@@ -103,6 +110,7 @@
             Country info:nb: Country info:nb
             Country info:cs: Country info:cs
             Country info:ru: Country info:ru
+            Country info:nl: Country info:
```
ohare93, line 113: Missing "nl" here

ohare93 (commented):
Looks good! Just one small catch. I will do a test run of it too :+1:

ohare93:
Test run successful :tada: The Dutch deck looks great! :grin: (from a cursory glance!)

WouterVdS:
Fixed the missing 'nl'. And thanks!

src/data/country_info.csv
```diff
@@ -77,24 +77,24 @@ Aegean Sea,,Auch Ägäis.,,,,,
 Balkan Peninsula,,Auch Balkan.,,,Også kalt Balkanhalvøya.,,
 Caspian Sea,,Auch Kaspisee.,,,,,
 Arabian Sea,,,,,Også kalt Arabiahavet.,,
```
axelboc, line 79: I see no changes on these lines. Should a coma be added at the end of each of them?

axelboc (approved):
Thanks @WouterVdS (also for documenting the Russian translation in the `README`)!! ❤️

src/data/country_info.csv
ohare93: Comma added because a new column was added, I would expect.

src/headers/desc.html
```diff
@@ -8,7 +8,7 @@
...
 
-The deck is available in <a href="https://github.com/axelboc/anki-ultimate-geography#other-languages-and-versions">five languages</a>: <b>English</b>, <b>German</b>, <b>Spanish</b>, <b>French</b>, <b>Norwegian</b> and <b>Czech</b>. An <a href="https://github.com/axelboc/anki-ultimate-geography#other-languages-and-versions"><b>extended version</b></a> is also available in each language. To help with memorisation and provide context while learning, some notes include extra information such as similar flags, governance information, alternative country names, etc.
+The deck is available in <a href="https://github.com/axelboc/anki-ultimate-geography#other-languages-and-versions">seven languages</a>: <b>English</b>, <b>German</b>, <b>Spanish</b>, <b>French</b>, <b>Norwegian</b>, <b>Czech</b>, <b>Russian</b> and <b>Dutch</b>. An <a href="https://github.com/axelboc/anki-ultimate-geography#other-languages-and-versions"><b>extended version</b></a> is also available in each language. To help with memorisation and provide context while learning, some notes include extra information such as similar flags, governance information, alternative country names, etc.
```
aplaice, line 11: Since I'm an idiot, and had not updated the number of languages after adding Czech, the number is wrong (we now have eight languages)!

src/data/country_info.csv
WouterVdS: Correct, I will add them. Rebasing on the Russian translation made me merge all the .csv files manually, which introduced this bug!

src/data/capital_info.csv
```diff
@@ -1,26 +1,26 @@
...
+Switzerland,Switzerland has no official capital; Bern is the de facto capital.,Die Schweiz hat keine offizielle Hauptstadt; Bern ist der Regierungssitz.,Suiza no tiene capital oficial; Berna es la capital de facto.,La Suisse n'a pas de capitale officielle ; Berne est la capitale de fait.,Sveits har ingen offisiell hovedstad. Bern er regjeringssete.,"Švýcarsko nemá žádné oficiální hlavní město, Bern je de facto hlavní město.",Швейцария не имеет официальной столицы; Берн является фактической столицей.,Zwitserland heeft geen officiële hoofdstad maar Bern vervult deze rol in de praktijk en wordt aangeduid als bondsstad
```
aplaice, line 7: To be extraordinarily pedantic, for consistency, there should be a full-stop at the end here. (Not that it obviously matters; it just looks slightly nicer :))

aplaice (approved):

src/data/country_info.csv
ohare93: Not a problem with Csvs anyways, the comma count does not need to sync up, especially if it's just less than the headers. As shown by the fact that I completed an output with no issues :grin:

aplaice:
Extreme-pedantry and an error-caused-by-me aside, this looks amazing! Thanks so much!

src/data/capital_info.csv
WouterVdS: Good spot! Fixed it! I'd be ashamed if that period wasn't there ;)

WouterVdS:
> Extreme-pedantry and an error-caused-by-me aside, this looks amazing! Thanks so much!

Happy to contribute! Wrote a small python script to open a Google search for the list of English countries, one every eight seconds. This way I could follow along in the .csv while Google showed me the country info in my own language!

aplaice:
> Happy to contribute! Wrote a small python script to open a Google search for the list of English countries, one every eight seconds. This way I could follow along in the .csv while Google showed me the country info in my own language!

That's pretty clever!

src/data/country_info.csv
axelboc: Might be a good idea to enforce it somehow, as it could be a problem for the next translation that comes along.

ohare93: I believe running a source_from_anki will write to the csvs correctly, though you may need to manually edit something so that there is an actual difference in the data that requires a write :thinking: hmm, worth thinking about on my end, I suppose

PR #384, opened 2021-02-02
Describe and apply full stop conventions

### LINKED ISSUE #383: Use of full stops at the ends of descriptive fields
We currently have a cross-language, unofficial convention that the `capital_info` and `country_info` fields end in full stops, while the `capital_hint` and `flag_similarity` ones don't.

I assume that it's because the former group at least sometimes contains full sentences, while the latter are always phrases.

However, most of the `country info`s are also simple phrases (e.g. `Constituent country of X`, `Island of X` or `Also known as X` (AFAIK, in English, if my grammar knowledge is correct, past participles don't count as verbs for the purpose of determining what is a sentence or clause)). The only "country info" that is actually a sentence that I could find, at least in English, was for the Sea of Japan. In contrast, almost all of the `capital info`s are sentences (e.g. `Cetinje is an honorary capital.`). Hence:

1. Should we revise the (currently unofficial) convention? 

    For instance, change the rule such that all phrases do not end in a full stop, and all sentences do, irrespective of which field they're in. In practice, that'd result in most of the `country info`s and a couple `capital info`s losing their trailing full stops. However, would that not look ugly/inconsistent (even if it'd be "grammatically correct")?

2. Should we document the conventions in `CONTRIBUTING.md`?

3. Are the rules about what constitutes a complete sentence different in different languages? (For example, there are some languages where "to be" can be omitted and only be "implicitly" present. (Approximately, "This table." means "This is a table.").) Hence, would that affect the dividing line between phrases and sentences, for that language?

<hr/>

(Spun out from #382.) (Hmm, I automatically added a full stop, despite it not being a sentence...)

<hr/>

This is obviously extraordinarily minor, but getting such stylistic issues right (and consistent) makes the deck slightly nicer to use!

aplaice:
<extreme_pedantry>

They were mostly applied in all languages.

(Apart from Swedish.)

Fix #383.

</extreme_pedantry>

(It should probably wait until after the 4.1 release, just in case — it's highly unlikely to mess anything up, but one never knows.)


CONTRIBUTING.md
```diff
@@ -370,6 +391,16 @@ The following guidelines apply to map images:
...
+
+By convention and for consistency, we use a full stop at the end of the `capital info` and `country info` fields and no full stop for the `capital hint` and `flag similarity` fields, in all languages, unless there are strong typographic or grammatical reasons against this in the given language. (No such language is currently in the deck.) The justification for this convention, as described in [#383][ref383] is that:
+
+1. The former two fields contain sentences (as opposed to phrases) relatively frequently, while the latter two almost never (exceptions include `(No) es un país soberano`). For consistency, and to avoid missing full stops when they are grammatically needed, it's useful for all the fields of the same type to use the same convention.
+2. It helps visually differentiate the former two "informative" fields, from the latter two "practical" ones.
+3. The former two fields, unlike the latter two, sometimes contain multiple sentences or multiple phrases (e.g. `Auch Nur-Sultan. Früher Astana.`), which need to be delimited with full stops, so always ending these sentences and phrases with a full stop makes sense.
+
```
axelboc, line 403: Awesome! I would just leave out the justifications. If people want to know more, they can look at the issue. You can move the list to a comment on the issue if you'd like, but since the issue isn't too long, I don't think it's needed.

CONTRIBUTING.md
```diff
@@ -370,6 +391,16 @@ The following guidelines apply to map images:
 - The style (colours, strokes, etc.) should match that of existing maps in the deck (note that water bodies use a different style than countries).
 - For small islands or archipelagos, the map should include a zoom box to facilitate identification.
 
+### Field style
+
+As mentioned above, we strive to keep the descriptive fields (`capital info`, `capital hint`, `country info`, `flag similarity`) as concise as possible. For instance, where the subject of the description is the same as the subject of the card, we use a truncated phrase instead of a full sentence — e.g. `Also known as Kiev.`, rather than `Kyiv is also known as Kiev.`.
+
+By convention and for consistency, we use a full stop at the end of the `capital info` and `country info` fields and no full stop for the `capital hint` and `flag similarity` fields, in all languages, unless there are strong typographic or grammatical reasons against this in the given language. (No such language is currently in the deck.) The justification for this convention, as described in [#383][ref383] is that:
+
```
axelboc, line 399: In previous paragraphs, we use italic for field names, e.g. _Capital info_.

CONTRIBUTING.md
```diff
@@ -370,6 +391,16 @@ The following guidelines apply to map images:
 - The style (colours, strokes, etc.) should match that of existing maps in the deck (note that water bodies use a different style than countries).
 - For small islands or archipelagos, the map should include a zoom box to facilitate identification.
 
+### Field style
+
+As mentioned above, we strive to keep the descriptive fields (`capital info`, `capital hint`, `country info`, `flag similarity`) as concise as possible. For instance, where the subject of the description is the same as the subject of the card, we use a truncated phrase instead of a full sentence — e.g. `Also known as Kiev.`, rather than `Kyiv is also known as Kiev.`.
+
```
axelboc, line 397: Nitpicking here, but I would just use double quotes for the example sentences, rather than back quotes.

CONTRIBUTING.md
```diff
@@ -370,6 +391,16 @@ The following guidelines apply to map images:
 - The style (colours, strokes, etc.) should match that of existing maps in the deck (note that water bodies use a different style than countries).
 - For small islands or archipelagos, the map should include a zoom box to facilitate identification.
 
+### Field style
```
axelboc, line 394: How about "Writing style"? Good call adding a new section, we can add more stylistic guidelines here in the future, if needed.

axelboc (approved):
Nice work finding all the missing full stops! 😁 Not as many as I thought, actually!

axelboc:
> (It should probably wait until after the 4.1 release, just in case — it's highly unlikely to mess anything up, but one never knows.)

Yeah, you're right, not worth rebuilding/reuploading all the decks.

axelboc:
Oh and thanks for updating the table of contents ! ❤️ 😄

aplaice:
Thanks for the nitpicking! Nitpicking guidelines deserve extra-thorough nitpicking! :)

I think that I've applied all your suggestions!

PR #385, opened 2021-02-06
Remove outdated information in LICENSE.md

aplaice:
All images are now properly sourced.

I don't think that we'll ever want to add images of dubious provenance, so we don't need the removed sentence.

(I accidentally noticed, now — we've had only properly sourced images for a while now.)

Also, hooray on the release! :) Thanks for doing it! You didn't mention yourself, in the release notes, but your care, guidance and attention to detail are essential for the well-being of the deck!


axelboc:
Awww shucks! 😊 😄

axelboc (approved):

Issue #386, opened 2021-02-06
Add more seas with maps


axelboc:
The new inclusion rules for seas, discussed in #346 and added in #379 are ready to be put into application! 🎉 

The limits to use are the ones defined by the IHO in its [2002 draft of _Limit of Oceans and Seas_](http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf).

### Seas to be added

- [ ] [Andaman Sea](https://en.wikipedia.org/wiki/Andaman_Sea)
- [ ] [Arafura Sea](https://en.wikipedia.org/wiki/Arafura_Sea)
- [ ] [Baffin Bay](https://en.wikipedia.org/wiki/Baffin_Bay)
- [ ] [Banda Sea](https://en.wikipedia.org/wiki/Banda_Sea)
- [ ] [Barents Sea](https://en.wikipedia.org/wiki/Barents_Sea)
- [ ] [Beaufort Sea](https://en.wikipedia.org/wiki/Beaufort_Sea)
- [ ] [Bering Sea](https://en.wikipedia.org/wiki/Bering_Sea)
- [ ] [Bismarck Sea](https://en.wikipedia.org/wiki/Bismarck_Sea)
- [ ] [Celebes Sea](https://en.wikipedia.org/wiki/Celebes_Sea)
- [ ] [Chukchi Sea](https://en.wikipedia.org/wiki/Chukchi_Sea)
- [ ] [East China Sea](https://en.wikipedia.org/wiki/East_China_Sea)
- [ ] [Great Australian Bight](https://en.wikipedia.org/wiki/Great_Australian_Bight)
- [ ] [Greenland Sea](https://en.wikipedia.org/wiki/Greenland_Sea)
- [ ] [Gulf of Aden](https://en.wikipedia.org/wiki/Gulf_of_Aden)
- [ ] [Gulf of Alaska](https://en.wikipedia.org/wiki/Gulf_of_Alaska)
- [ ] [Gulf of Guinea](https://en.wikipedia.org/wiki/Gulf_of_Guinea)
- [ ] [Gulf of Saint Lawrence](https://en.wikipedia.org/wiki/Gulf_of_Saint_Lawrence)
- [ ] [Ionian Sea](https://en.wikipedia.org/wiki/Ionian_Sea)
- [ ] [Java Sea](https://en.wikipedia.org/wiki/Java_Sea)
[...]

PR #388, opened 2021-02-06
Add first set of sea maps

aplaice:
_(For release notes: part of #386 + panning of the Yellow Sea map)_

These are the ones that didn't seem to differ from the IHO 2002 maps at all. Many others differ very slightly (e.g. one of the limits is from the northern rather than the southern end of an island...).

`src/data` and `sources.csv` not changed yet.

Please feel leave feedback regarding panning/zooming in etc.! (Panning/zooming is easy to change!)

## Banda Sea

http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=135

## Barents Sea

http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=205

## Celebes Sea

http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=147

## East China Sea

http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=162

Possibly zoom in? (But the context is more important and zooming in more won't really provide much value. Compare [the Yellow Sea](https://github.com/axelboc/anki-ultimate-geography/blob/master/src/media/maps/ug-map-yellow_sea.png) and [the South China Sea](https://github.com/axelboc/anki-ultimate-geography/blob/master/src/media/maps/ug-map-south_china_sea.png).

## Gulf of Alaska

http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=176

## Gulf of Guinea

http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=25

## North Sea

http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=16

Northern limit?

## Sea of Okhotsk

http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=171

## Timor Sea

http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=98

(Includes Joseph Bonaparte Gulf, but I think that that's correct.)


axelboc:
- Banda Sea ✔️ 
- Celebes Sea ✔️
- Gulf of Alaska ✔️
- Gulf of Guinea ✔️
- Sea of Okhotsk ✔️
- Timor Sea

It is indeed correct to include the Joseph Bonaparte Gulf. I would perhaps just pan to the West a little.

- Barents Sea

Very slight discrepancy with IHO: the western limit should go around the little island North of Norway:

image 

- East China Sea

How about panning to the West a little, so Southeast Asia and the Philippines are a bit more visible? Also, I feel like the eastern limit is a tiny, tiny bit off -- it should pass a bit East of one of the small islands:

image

- North Sea

Maybe pan to the East a tiny bit? Regarding the limits, there are quite a few subtle differences with the IHO:

- Northwest limit should be a straight line between the two islands instead of curving to the tiny island in the middle: 

image

- Northern limit should start at the tip of the northern most island):

image

- Northeastern limit should include both fjords:

image

- Waterways in the tip of Denmark should not be included:

image

- Southern limit should go past the chain of islands up to the entrance of the main bay of the Netherlands:

image

- ... and should not include the other bay near Belgium:

image

---

Those are all really minor details, so I'm not fussed if they're too hard to fix. 😉 They look awesome already 💯

aplaice:
## Timor Sea

Panned slightly west.



## East China Sea

Panning changed



## Yellow Sea

I've also changed it, since it probably should have the same panning as the East China Sea (and IMO it looks better now, compared to how I had previously panned it...).

### Old



### New



## North Sea

Panning and some of the limits changed.



## Not applied (yet?)

[ ] Barents Sea (western limit)
[ ] East China Sea (eastern limit)
[ ] North Sea (northwest limit)
[ ] North Sea (northern limit)
[ ] North Sea (waterways in the tip of Denmark)
[ ] North Sea (bay near Belgium)

(They're generally trickier, since they require the adjusting of the shapefiles, which I'm relatively hesitant to do, since doing it fully correctly is tricky, while doing it just-about-good-enough is still relatively time consuming, but feels "wrong" (I trust the shapefiles from Natural Earth Data to represent something, while I don't quite trust my own shapefiles which would neither be NE, nor quite IHO 2002 ones...

However, for some of the other cases where the discrepancies are greater, I might need to edit shapefiles, anyway, so if I get "into the swing of things" I might come back to some of these. (I've used my own shapefiles for the Bering Strait, and it wasn't *that* bad.)

If I do, then some of the other maps (e.g. the Norwegian Sea, which borders both the Barents Sea and the North Sea) might need to be updated, as well.)

axelboc:
No pressure, these are great as they are and already tremendously beneficial to the deck! 💯

axelboc:
I'm going to move all the tiny discrepancies that require modifying the shapefile into a separate issue. Let's get all those beautiful maps merged in, shall we!?

axelboc:
Oh right, `src/data` and `sources.csv` still need to be updated, sorry. (Though for the sources, you can put `TBC` or something for now, I reckon.)

So my point is: happy to merge all your maps in their current state whenever you're ready. I'll open up the issue with the discrepancies once the PR is merged.

aplaice:
I think I'm done (with this set of maps). For review purposes `31e5ffb` is the master-rebased version of the previous last commit.

<hr/>

I've added the English, German, Spanish and French translations but none of the others.

I'm ignoring the French name variation "Golfe de l'Alaska", as being similar to the main variant and the old German name "Meerbusen von Guinea" since quickly googling it, the only non-Wiki results appear to be colonial-era maps.

axelboc (approved):
Awesome!

I've added the remaining translations. A few of the article titles on Russian Wikipedia follow the homonymy scheme - e.g. _Alaska (gulf)_, so I had to make some adjustments.

I think it's better to have potentially incorrect translations than no translations at all, as it's easier to fix something that exists than to realise that something is missing. That being said, I didn't look at alternative names at all, they're above my pay grade :D

aplaice:
> I've added the remaining translations.

Wow! Thanks! You're right that some translation is better than no translation.

I wish we had a way of marking translations as "to be reviewed". OTOH by the law of the internet the best way to get an answer is to make a (potentially incorrect) statement, so we'll probably get feedback anyway. :)

axelboc:
The minor insconsistencies with the IHO are now documented in a discussion thread: https://github.com/axelboc/anki-ultimate-geography/discussions/412

PR #396, opened 2021-02-13
Fix guids for other languages

### LINKED ISSUE #391: Notes in translated decks have new GUIDs
See https://github.com/axelboc/anki-ultimate-geography/discussions/390#discussioncomment-346750 for context, and https://github.com/Stvad/CrowdAnki/issues/6 for root cause.

This issue is to "consider whether we should recommend using the _Do Not Move Existing Cards_ option and whether we should add guidance/reassurance about the creation of a "duplicate" deck" on first CrowdAnki import.

---

**EDIT** the duplicated deck issue is more severe than initially thought -- see https://github.com/axelboc/anki-ultimate-geography/issues/391#issuecomment-777310776 below.

ohare93:
Fixes #391 ... mostly :cry: 

Changes:
 - Add new Guid.csv which contains the correct guids for each language, as they were previously 
     - New languages that did not exist before have had new guids generates for every note
     - New notes that did not exists in any language before have had new guids generated too
     - I used the [this Gist](https://gist.github.com/ohare93/900706a79e6a1879808d02152c0e3be5) to copy the old guids into the right column :+1: though there were issues with certain entries that have changed in name slightly, that I had to copy by hand (so please double check these ones!)
         - German: Oceania and St Lucia
         - Spanish: Romania
 - Change the recipe to handle these new guids
     - I have only updated the source_to_anki for now, the other direction may fail at the moment

I have released a new version of Brain Brew (v0.3.1) for this fix (very minor one line fix to stop throwing an error) and updated the version in this repo, so please remember to update your local before running it.

It turns out the guid generation is a bit more tangled than I would like it, as it expects just one guid per import. I have disabled it in UG for now. I will make a change to Brain Brew next week that will turn Guid generation into it's own separate task, which one can specify the specific column**s** to do it on :+1: I thought it was better to get this fix in so the deck assets can be fixed immediately, rather than spend time on this (which I do not have until Wednesday)

# Big Issue!
It turns out that **not all guids were different!** :scream: At least in the [Release 4.1](https://github.com/axelboc/anki-ultimate-geography/releases/tag/v4.0) assets:
  - Spanish and French are identical
  - English and Norwegian are identical

This is a big issue :sob: we will have to make a decision either now or in the future which will cause _someone_ to lose progress on updating :disappointed: 

I have no idea how this happened :thinking: maybe a bug in Anki-DM that went unnoticed?


ohare93:
Link to the Brain Brew change: https://github.com/ohare93/brain-brew/commit/3f8ef551d624919404640d4a283ff88de294d901 very minor, as I said.

aplaice:
Thanks for doing this! I fully agree with all the decisions you made and will look through the details of the changes, now!

> I thought it was better to get this fix in so the deck assets can be fixed immediately, rather than spend time on this (which I do not have until Wednesday)

Yes, definitely, it's not urgent! 

<hr/>

> Big Issue!

> It turns out that not all guids were different! scream Spanish and French appear to be identical in their guids. At least in the Release 4.1 assets.

Wow, you're right — they're all identical, also in my local (old) copy of the `anki-dm` builds.

<details>

```bash
diff <( jq < Ultimate\ Geography_es/Ultimate\ Geography_es.json '.notes[].guid' ) <( jq < Ultimate\ Geography_fr/Ultimate\ Geography_fr.json '.notes[].guid')
```

</details>

> I have no idea how this happened  maybe a bug in Anki-DM that went unnoticed?

Presumably; I wonder whether it's linked to the "numerical values"/unicode representations of the letters of `fr` and `es` adding up to the same number.

> we will have to make a decision either now or in the future which will cause someone to lose progress on updating disappointed

Yeah :(. I think that it can wait, though.

aplaice (approved):
The copied GUIDs (as reported by `csvtool`/`jq`, where I assume that the order of notes hasn't changed; I manually checked the cases you had listed as problematic; it's highly unlikely that both your approach and my lazy one would be wrong) from 4.0 are all correct, the new decks build correctly, the copied GUIDs in the generated decks are also correct (again as reported by `jq`) and testing importing the generated NB deck over the 4.0 NB deck results in only 7 new cards (exactly as expected).

I'm not certain about keeping the "main"/English `guid` column in `src/data/main.csv` in addition to having it in `src/data/guid.csv`, but on the whole I think it's the right choice, and in any case, it can be trivially changed later.

Thanks again!

ohare93:
> I'm not certain about keeping the "main"/English `guid` column in `src/data/main.csv` in addition to having it in `src/data/guid.csv`, but on the whole I think it's the right choice, and in any case, it can be trivially changed later.

Oh, that was unintentional, I meant to remove that. Curious that Brain Brew did not complain about duplicate columns :thinking: it probably ignored the english guid column in guid.csv. I will look into that! :sweat_smile:

ukanuk:
> This is a big issue 😭 we will have to make a decision either now or in the future which will cause _someone_ to lose progress on updating 😞

Would it be possible to update CrowdAnki to enable migrating card GUIDs? You'd upload a file with old>new GUID info, and it would directly modify the Anki database accordingly. I could see this causing sync issues - but presumably they'd be resolved with a full upload to AnkiWeb immediately after the modification.

ohare93:
> Would it be possible to update CrowdAnki to enable migrating card GUIDs? You'd upload a file with old>new GUID info, and it would directly modify the Anki database accordingly. I could see this causing sync issues - but presumably they'd be resolved with a full upload to AnkiWeb immediately after the modification.

Possible, totally! But difficult to get right, and would take some effort to do. Very bad if there's a bug or oversight in _that_ code :scream:

aplaice:
> Would it be possible to update CrowdAnki to enable migrating card GUIDs? You'd upload a file with old>new GUID info, and it would directly modify the Anki database accordingly. I could see this causing sync issues - but presumably they'd be resolved with a full upload to AnkiWeb immediately after the modification.

Yeah, I've been thinking about something like that, but I don't think that it should be part of `CrowdAnki`, as such — like @ohare93 notes, there's scope for catastrophic error, so it should be a separate add-on in order that users know exactly what they're asking for. (I haven't had time to do any proper coding, though.)

ohare93:
Removed the Guid column from Main.csv :+1: still builds fine, no worries.

ohare93:
Oh... I have just noticed: **Norwegian Guids are also the same as the English Guids** :facepalm: I will update my original comment :disappointed:

axelboc:
Excellent, thank you!! I'll update the release later today.

axelboc:
Just had a thought: isn't the fact that the CrowdAnki UUIDs of all the decks are the same preventing anyone from using multiple decks at all?

aplaice:
> Just had a thought: isn't the fact that all the decks have the same CrowdAnki UUID preventing anyone from using multiple decks at all?

The UUIDs all being the same shouldn't matter. The deck UUID being the same would simply result in all the notes/cards being dumped in the same deck (which is mildly annoying (or useful, depending on your perspective) but not that important), the deck config UUID being the same would have no discernable effect, while the note model UUID  being the same would also have no effect, since our note models aren't (yet! :)) localised.

However, that's actually a good point regarding upgrading! (My test of upgrading the NB deck, mentioned above, was useless due to the EN and NB GUIDs being the same, like @ohare93 noticed.)

With `anki-dm`, the set of the three UUIDs was different for each deck (or for each group of decks, (EN, NB), (FR, ES), DE).

With the UUIDs becoming the same for all languages (i.e. changing for FR, ES and DE), upgrade will still occur without data loss. 

However, due to the deck UUID changing, a "duplicate" deck will be created (with the notes moved there (i.e. no duplicate notes)) and due to the note model uuid changing, there will be a scary (but harmless — unless you mess with any of the options) note model migration dialog (`Current note type: Ultimate Geography`, `New note type: Ultimate Geography-xxxxx`).

(That's what I've now noticed with the DE deck, and is what one should expect.)

The former issue (duplicate deck) is harmless, but it's annoying as it's the same symptom as when real problems (i.e. all the notes being duplicated) occur. (It's what made us ignore the issue here, at first — what was reported on reddit regarding the English deck, was harmless, the issue reported on GitHub regarding the French deck was serious.)

I don't think that we should fiddle with making the UUIDs per-deck, as it would hugely over-complicate all the YAMLs (and IMO it makes sense for the deck config and note model UUIDs to be the same), but we probably should mention/warn users about what might happen.

I'll try to think about the best/simplest diagnostics for users to check that upgrading has gone well.

:scream 

<hr/>

> Makes me think that there is either a low number of people using the translated decks and/or a low number of people trying to use multiple decks

Hopefully the latter! :)

aplaice:
> but we probably should mention/warn users about what might happen.

> I'll think about the best/simplest diagnostics for users to check that upgrading has gone well, and what would be a simple warning message, today.

This is overly verbose, but perhaps something like this:

> Note that if you're upgrading your Spanish, French or German Ultimate Geography deck from 4.0 to 4.1, you will see a "Change Note Type" pop-up, which will tell you that the note type is being changed from "Ultimate Geography" to "Ultimate Geography-xxxx" where xxxx is a random alpha-numeric string, (or from "Ultimate Geography [Extended]" to "Ultimate Geography [Extended]-xxxx"). This is due to technical changes in how we generate the decks and is harmless. You shouldn't have to change any of the default options, in this popup.

> You may also see a duplicate "Ultimate Geography" deck on your Anki main screen. If you're worried, please make sure that the "new" deck (the one that says v4.1 in its description) contains 975 (or 1508 if you're using the extended) cards, while the "old" deck contains no or almost no cards (there may be some cards from the legacy deck, now no longer in the main deck). You can check the number of cards in a deck by selecting the deck and clicking on "Stats" at the top (or pressing <kbd>t</kbd>). If that's not the case, please report it [here](https://github.com/axelboc/anki-ultimate-geography/issues)! You can remove the "old" deck, first optionally moving any cards in it into the new deck.

<details>

```markdown
Note that if you're upgrading your Spanish, French or German Ultimate Geography deck from 4.0 to 4.1, you will see a "Change Note Type" pop-up, which will tell you that the note type is being changed from "Ultimate Geography" to "Ultimate Geography-xxxx" where xxxx is a random alpha-numeric string, (or from "Ultimate Geography [Extended]" to "Ultimate Geography [Extended]-xxxx"). This is due to technical changes in how we generate the decks and is harmless. You shouldn't have to change any of the default options, in this popup.

You may also see a duplicate "Ultimate Geography" deck on your Anki main screen. If you're worried, please make sure that the "new" deck (the one that says v4.1 in its description) contains 975 (or 1508 if you're using the extended) cards, while the "old" deck contains no or almost no cards (there may be some cards from the legacy deck, now no longer in the main deck). You can check the number of cards in a deck by selecting the deck and clicking on "Stats" at the top (or pressing <kbd>t</kbd>). If that's not the case, please report it [here](https://github.com/axelboc/anki-ultimate-geography/issues)! You can remove the "old" deck, first optionally moving any cards in it into the new deck.
```

</details>

Feel free to shorten/edit mercilessly/ignore!

PR #403, opened 2021-02-21
Rewrite installation and upgrade docs in README

### LINKED ISSUE #398: Improve Minor upgrade instructions
Thanks for maintaining this great deck. I'm trying to update from 4.0 to 4.1
When performing a minor upgrade i follow the instructions and use crowd anki.
But im not sure what should be ticked or not. 

In my screenshot is the default selection and the instructions on the right. 

Screenshot 2021-02-14 173709

Secondly there's no mention of a new deck being created. I have "ultimate georgaphy_2" what's going on here?
image

axelboc:
Fix #398. Hopefully it's a lot clearer:

- The _Getting started_ section now explains how to install the deck with CrowdAnki. I no longer mention the APKG install so that people don't end up in a situation where they have a duplicated deck on upgrade. Also, the steps are now the same when installing and upgrading, which is much simpler to understand.
- I document a few common upgrade scenarios in the _Upgrading_, as you'll see. Am I missing any?
  - Should we add scenarios for when people change the templates/styles/fields?
  - Is the _Standard to extended_ section still accurate, or could we provide people with a better solution?
- Should I add a note about importing two decks in two different languages somewhere?

---

**EDIT**

I've added a new section called _Customising the deck_ that goes through a number of changes that users might be interested in making to their local UG decks. It explains how some changes may affect or be reversed by future upgrades, how to go about preserving the changes when possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,16 @@ The deck is available in **English**, **German**, **Spanish**, **French**, **Nor
   - [Supplementary information](#supplementary-information)
   - [Custom study](#custom-study)
 - [**Getting started**](#getting-started)
-  - [Other languages and versions](#other-languages-and-versions)
 - [**Upgrading**](#upgrading)
-  - [Minor version](#minor-version)
+  - [First upgrade after APKG import](#first-upgrade-after-apkg-import)
+  - [Keeping deck customisations](#keeping-deck-customisations)
+  - [Levelling up from standard to extended](#levelling-up-from-standard-to-extended)
   - [Major version](#major-version)
 - [**Deck structure**](#deck-structure)
+- [**Customising the deck**](#customising-the-deck)
+  - [Changes that are or can be preserved](#changes-that-are-or-can-be-preserved-)
+  - [Changes that get reverted](#changes-that-get-reverted-)
+  - [Changes that prevent upgrading](#changes-that-prevent-upgrading-)
 
 ## Features
 
@@ -121,15 +126,9 @@ If you're in this situation and wish to upgrade, proceed as follows:
 1. You should end up with two decks: your original deck called "Ultimate Geography" and a new, duplicated deck called "Ultimate Geography_2". Unless you created some notes of your own, every card in your original deck should have moved automatically to the new deck and your original deck should now be empty. You can verify this in Anki's [card browser](https://docs.ankiweb.net/#/browsing?id=browsing).
 1. To get back to having a single deck, move any of your own cards out of the original deck and delete it. Then, rename the new deck to "Ultimate Geography". From here on, future upgrades will be seamless.
 
-### Keeping any moved cards where they are
+### Keeping deck customisations
 
-If you've moved some of the cards out of the default _Ultimate Geography_ deck and into another deck, by default CrowdAnki will move those cards back on import.
-
-To prevent this behaviour and update the existing cards in place, follow the steps in the [_Getting started_](@getting-started) section but in the last step, make sure to tick the _Do Not Move Existing Cards_ checkbox in the _CrowdAnki Import Settings_ dialog box.
-
-### Keeping your own cards
-
-Unless otherwise mentioned in the release notes, if you've added your own cards to the deck, you can be reassured that upgrading will not affect them whatsoever. If a duplicated deck is created on import for whatever reason, the cards you created will always remain in their original deck.
+If you've made, or are thinking of making, any changes to your _Ultimate Geography_ deck, check the [_Customising the deck_](#customising-the-deck) section below to see what will happen to them the next time you upgrade, and whether you need to take any steps to preserve them.
 
 ### Levelling up from standard to extended
 
@@ -140,17 +139,17 @@ Importing an extended deck on top of a standard deck may bring unexpected result
 
 If you'd rather keep your progress, proceed as follows:
 
-1. Perform the _Getting started_ steps exactly or as per the [_Cards moved into other deck_](cards-moved-into-other-deck) scenario.
+1. Perform the _Getting started_ steps exactly.
 1. Instead of a dialog box confirming that the import was successful, you'll see another dialog box titled _Change Note Type_. In the _Cards_ section, set _Change Flag - Country to:_ to "Flag - Country" (instead of "Country - Flag"), and set _Change Map - Country to:_ to "Map - Country" (instead of "Flag - Country").
 1. Check that the other cards and fields map correctly and click on _OK_ to start the import.
 
 ### Major version
 
-Major versions (e.g. `v3.0`) typically indicate that upgrading may lead to a loss of progress. This occurs, for instance, when cards are removed or when changes are made to the structure of the deck.
+Major versions (e.g. `v3.0`) typically indicate that upgrading may lead to a significant loss of progress, or to the loss of [some of the customisations](#customising-the-deck) you may have made to the deck that would normally be preserved. This can occur, for instance, when [significant changes](https://github.com/axelboc/anki-ultimate-geography/releases/tag/v3.0) are made to the structure of the deck, or when [lots of cards](https://github.com/axelboc/anki-ultimate-geography/releases/tag/v4.0) are removed.
 
 > For more information about versioning and the differences between minor and major, please refer to the [_Versioning_](CONTRIBUTING.md#versioning) section of the _CONTRIBUTING_ guide.
 
-As for any release, the release notes of a major version will tell you how to upgrade, or point you to a page that does. In some cases, there may even be a way to keep your progress! However, this is rarely straightforward. If you don't mind losing your progress, the simplest way to upgrade to a major version is to perform a "clean import":
+Read the release notes carefully to know if the loss of progress/customisation actually concerns you. If it does, the release notes will tell you how best to upgrade and may even offer you ways to preserve or mitigate your losses. Worst-case scenario, you may have no other choice but to accept the losses and perform a "clean import" as follows:
 
 1. Open Anki and make sure your devices are all synchronised.
 1. Delete the `Ultimate Geography` deck.
@@ -160,8 +159,6 @@ As for any release, the release notes of a major version will tell you how to up
 1. You can now follow the steps of the [_Getting started_](#getting-started) section to install the major version.
 
 ## Deck structure
-
-> Beware that changing the structure of the deck may **prevent you from upgrading it** without loss of progress.
 
 The notes are based on a **note type** called _Ultimate Geography_, which defines **eight fields**: _Country_, _Country info_, _Capital_, _Capital info_, _Capital hint_, _Flag_, _Flag similarity_ and _Map_.
 
@@ -174,3 +171,84 @@ Every note is tagged. The tags, which are listed below by category, are prefixed
 - Type - `UG::Sovereign_State`, `UG::Oceans+Seas`, `UG::Continents`
 - Continent - `UG::Africa`, `UG::Asia`, `UG::Europe`, `UG::North_America`, `UG::Oceania`, `UG::South_America`
 - Region - `UG::Caribbean`, `UG::Middle_East`, `UG::Southeast_Asia`, `UG::European_Union`, `UG::Mediterranean`, `UG::East_Africa`, `UG::West_Africa`
+
+## Customising the deck
+
+Interested in making changes to the deck in Anki? Here's what you need to know.
+
+### Changes that are or can be preserved 💖
+
+- **Adding your own cards and media files**
+
+  Go wild. 🌊 Upgrading will not affect any new cards you create, or any new media files you add.
+
+  If a duplicated deck is [created on import](#first-upgrade-after-apkg-import), your own cards will stay in their original deck.
+
+- **Moving cards into another deck**
+
+  Some people prefer to combine multiple decks into one to review them together.
+
+  By default, CrowdAnki will move all the cards back into the _Ultimate Geography_ deck on import. To prevent this behaviour and update the existing cards in place, follow the steps in the [_Getting started_](@getting-started) section but in the last step, make sure to tick the _Do Not Move Existing Cards_ checkbox in the _CrowdAnki Import Settings_ dialog box.
+
+  > Note that if you had previously deleted the _Ultimate Geography_ deck, it will be recreated and may contain new cards.
+
+- **Changing the deck's options**
+
+  Deck options are, for instance, the number of new cards per day, the maximum number of reviews per day, etc.
+
+  When importing the deck for the first time, CrowdAnki creates a new "option group" called _Ultimate Geography_ (unless it already exists). If you've made any changes to this group's options, they will be reverted on import.
+
+  To avoid losing your options, create your own option group and give it a distinctive name. After you upgrade, you'll just have to switch the deck back to this option group.
+
+- **Customising the templates and styles**
+
+  Any edits you make to the built-in templates of the _Ultimate Geography_ note type will be reverted on upgrade. Same goes with the CSS styles that are shared by all the templates.
+
+  One way to workaround this and preserve your changes is to  first duplicate the note type and switch all the cards to the new note type in the Anki browser. You can then make your template and style changes in the duplicate note type without affecting the original note type. The next time you upgrade, all the cards will switch back to the original note type, but you can easily switch them back to your customised note type in the Anki browser.
+
+  > Note that you may have to update your custom templates and styles afterwards to stay in line with the built-in templates.
+
+- **Adding or removing a template**
+
+  We're getting into dangerous territory here, but you should be able to make this work by duplicating the default note type as explained in the previous point, _Customising the templates and their styles_.
+
+  The difference is that, when you import the deck and/or when you switch back the cards to your customised note type, you may see a dialog box asking you to match the templates of one note type to the templates of the other note type. (This same dialog box appears when importing the extended deck [on top of the standard deck](#levelling-up-from-standard-to-extended).)
+
+  As long as you map the templates that Anki asks you to match correctly, you should be fine. However, we strongly recommend that you try this out in a separate Anki profile first.
+
+### Changes that get reverted 😬
+
+The changes listed below will be lost the next time you upgrade and there's no other solution than to make the changes again manually after the upgrade.
+
+- **Renaming the deck or changing its description**
+
+  Upgrading always restores the deck's default name and description.
+
+- **Editing the content of a note**
+
+  Examples:
+
+  - adding a link to Wikipedia in the _Country info_ field of Curaçao;
+  - adding your own memoization hint to the _Flag similarity_ field of Poland;
+  - adding or removing a tag from a note;
+  - renaming a tag.
+
+  > If you correct a typo or a mistake, don't forget to [open an issue](https://github.com/axelboc/anki-ultimate-geography/issues/new) or a Pull Request so you don't lose your correction the next time you upgrade. 😉
+
+- **Deleting a card**
+
+  Upgrading will bring the card back. If you don't want to learn a card at all or if you want to delay learning it, consider:
+
+  - suspending or burying the card;
+  - creating a [filtered deck](#custom-study) that doesn't include the card;
+  - moving the card into an "icebox" deck and ticking the _Do Not Move Existing Cards_ the next time you upgrade, as explained in the previous section, under _Moving cards into another deck_.
+
+### Changes that prevent upgrading ⛔
+
+- **Adding a new field (population, currency, etc.)**
+
+  Adding a new field to the _Ultimate Geography_ note type will prevent you from upgrading the deck entirely. When you'll try to import it again, Anki will likely error and you'll risk losing your entire progress. Hopefully Anki, CrowdAnki, Brain Brew and UG will [one day](https://github.com/ohare93/brain-brew/issues/4#issuecomment-644975261) find a way to make this possible.
+
+  Note that the technique of duplicating the note type explained under _Customising the templates and styles_ will not work here. If you've added a new field and are wanting to upgrade, you have no choice but to first **remove the field** from the note type.
+
+  > If you really know what you're doing, you could try exporting your deck with CrowdAnki, carefully merging the JSON file of the new version of the deck into your exported deck's JSON file, and then importing your deck back into Anki ... but if you're capable of this, your skills would be put to much better use contributing to [CrowdAnki](https://github.com/Stvad/CrowdAnki) and [Brain Brew](https://github.com/ohare93/brain-brew/)! 😁
--- a/README.md
+++ b/README.md
@@ -93,19 +93,19 @@ In order to install and later upgrade _Ultimate Geography_ you'll need to first 
 
 1. Open Anki on your computer, go to the _Tools_ menu and select _Add-ons_.
 1. In the dialog box, click on _Get Add-ons..._ and paste in the code provided on [this page](https://ankiweb.net/shared/info/1788670778).
-1. Click on _OK_ to download the add-on, and then restart Anki.
+1. Click on _OK_ to install the add-on, and then restart Anki.
 
 You're now ready to install _Ultimate Geography_:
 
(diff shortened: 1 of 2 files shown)


aplaice:
> Is the Standard to extended section still accurate, or could we provide people with a better solution?

>  Importing an extended deck on top of a standard deck may bring unepected results. Instead, we recommend that you start learning the extended deck from scratch. To do so, either:

Importing an extended deck on top of a standard one may indeed bring unexpected (and without care undesired) results. However, with proper care, it can be done without issues:

1. Start the import the of the extended deck on top of the standard deck, as normal.
2. In the "CrowdAnki Import Settings" dialog, either tick the "Do not move existing cards" box, or not, as normal.
3. A "Change Note Type" dialog will appear, unlike normally. Here, the mapping between the existing Cards and the new Cards and between the existing Fields and the new Fields is defined. 
    Nothing needs to be changed in the "Fields" section, as the order of the Fields is the same in both note types (standard and extended).
    However, in the Cards section, you _**need** to set "Change Flag - Country to: Flag - Country" and "Change Map - Country to: Map - Country" (the "defaults"\* are mapping "Flag - Country" to "Country - Flag" and "Map - Country" to "Flag - Country", which is obviously wrong). 

\* The defaults are this way because of the order of the templates in the two note types. For instance, see the current [`recipes/source_to_anki.yaml`](https://github.com/axelboc/anki-ultimate-geography/blob/master/recipes/source_to_anki.yaml#L50-L67). (It was the same when using `anki-dm`.) Unfortunately, we can't easily rearrange the templates, so that the first four "standard" templates were also the first four "extended" templates, in the same order, because that would result in progress loss (or worse progress mixup) for everybody who was upgrading.

(FWIW I had upgraded from the standard to the extended deck in this way, ~ two years ago.)

README.md
```diff
@@ -87,52 +87,67 @@ Anki's [search feature](https://docs.ankiweb.net/#/searching), coupled with the
...
+
+### First upgrade after APKG import
+
+You may have initially installed _Ultimate Geography_ by importing an **APKG file**. You may have downloaded such a file from this repo or from the [deck's page](https://ankiweb.net/shared/info/2109889812) on AnkiWeb, as this used to be the recommended installation process.
+
+If you're in this situation and wish to upgrade, proceed as follows:
 
-> For more information about versioning and the differences between minor and major, please refer to the [_Versioning_](CONTRIBUTING.md#versioning) section.
+1. Perform the _Getting started_ steps exactly.
+1. You should end up with two decks: your original deck called "Ultimate Geography" and a new, duplicated deck called "Ultimate Geography_2". Every card in your original deck should have moved automatically to the new deck and your original deck should now be empty. You can verify this in Anki's [card browser](https://docs.ankiweb.net/#/browsing?id=browsing).
```
aplaice, line 121: > your original deck should now be empty

Should we mention something like "unless you created some notes of your own"?

README.md
```diff
@@ -87,52 +87,67 @@ Anki's [search feature](https://docs.ankiweb.net/#/searching), coupled with the
...
+If you've moved some of the cards out of the default _Ultimate Geography_ deck and into another deck, by default CrowdAnki will move those cards back on import.
 
-> If you've moved some of the cards out of the default _Ultimate Geography_ deck and into another deck, by default CrowdAnki will move those cards back on import. This behaviour [can be disabled](https://github.com/Stvad/CrowdAnki#configuration-settings).
+To prevent this behaviour and update the existing cards in place, follow the steps in the [_Getting started_](@getting-started) section but in the last step, make sure to tick the _Do Not Move Existing Cards_ checkbox in the _CrowdAnki Import Settings_ dialog box.
+
+### Standard to extended
+
+Importing an extended deck on top of a standard deck may bring unepected results. Instead, we recommend that you start learning the extended deck from scratch. To do so, either:
```
aplaice, line 132: Typo: "unepected results" → "unexpected results".

aplaice (approved):

aplaice:
> Should we add scenarios for when people change the templates/styles/fields?

There's no neat solution for this ([though there are workarounds, at least for the first two](https://github.com/axelboc/anki-ultimate-geography/discussions/400)), so perhaps?

> Should add a note about importing two decks in two different languages somewhere?

I think we should wait with mentioning it explicitly until we've disentangled French from Spanish and Norwegian from English.

> Am I missing any?

Perhaps the user having created their own cards, which obviously causes no issues?

<hr/>

Perhaps we could mention something to the effect of "for unusual upgrades and when you yourself have made changes other than those mentioned there are possible upgrade procedures or workarounds, as described here", linking to a Wiki page (so that anybody could add items there), which would, in turn link to multiple discussion pages, one per topic, similarly to what you had suggested regarding per-upgrade discussion pages.

(A Wiki page to emphasise that they're not "official"/recommended/that while we're willing to help, the user is responsible for whatever happens.)

The tricky standard→extended deck upgrade could also be mentioned in such a wiki and/or discussion page.

README.md
```diff
@@ -129,11 +129,17 @@ To prevent this behaviour and update the existing cards in place, follow the ste
 
 ### Standard to extended
 
-Importing an extended deck on top of a standard deck may bring unepected results. Instead, we recommend that you start learning the extended deck from scratch. To do so, either:
+Importing an extended deck on top of a standard deck may bring unexpected results. You should consider learning the extended deck from scratch. To do so, either:
 
 - remove the standard deck first by following the instructions in the [_Major version_](#major-version) section below, or
 - import the extended deck in a separate Anki profile.
 
+If you'd rather keep your progress, proceed as follows:
```
aplaice, line 137: Looks great! (Less verbose and more clear than what I wrote!)

axelboc:
I'll add a section for changes to fields, templates, etc. and point to the discussion. I'll keep things "high level" -- we can always point to other discussions later on with more detailed solutions.

> I think we should wait with mentioning it explicitly until we've disentangled French from Spanish and Norwegian from English.

Let me rephrase: should we add a warning to recommend to _not_ attempt importing two languages? 😁 

> Perhaps the user having created their own cards, which obviously causes no issues?

I guess a user's own cards could be affected if we made a radical change to fields or templates... To cover all our bases, I've added "Unless otherwise mentioned in the release notes, if you've added your own cards to the deck, you can be reassured that upgrading will not affect them whatsoever."

---

I like your idea of a wiki page, but I think we can manage without. I don't see a problem of having a few more unusual scenarios in the README (as long as they don't relate to specific languages/releases of the deck) -- if the solutions are too long, we can always point to the relevant issues/discussions. And for very obscure scenarios, I think once-off (but still searchable) issues/discussions are fine.

aplaice:
> Let me rephrase: should we add a warning to recommend to not attempt importing two languages? grin

I don't think it's worth it. It's a relatively niche situation and the README is long already. If somebody does do it, then they'll either have the deck in both languages or have their existing deck translated into the target language (other than the last seven notes added in 4.1, which will be created as new cards). Neither of these is catastrophic and if somebody becomes worried, they can ask here.

Since, once we disentangle EN/NB and FR/ES and upgrade everyone using CZ/NL/RU/SV from 4.1 to 4.1.1, having multiple languages will be explicitly (?) allowed, I'm not sure we want to confuse people by listing all the possible combinations of what can happen when they import two different languages.

We could mention something like "Importing two different languages is currently not explicitly supported, but nothing particularly bad is likely to happen if you do. Please see [here] for details or ask in the discussions section."?

axelboc:
I've spent some time trying all kinds of customisation scenarios, from renaming the deck to adding a new template, and I've documented their impact on upgrading and how people might go about preserving their changes. Let me know what you think!

axelboc:
Sorry for the force-pushed -- hopefully you hadn't started reviewing yet. I'm done now 😄

README.md
```diff
@@ -87,57 +92,74 @@ Anki's [search feature](https://docs.ankiweb.net/#/searching), coupled with the
...
+1. Check that the other cards and fields map correctly and click on _OK_ to start the import.
 
 ### Major version
 
-Upgrading to a major version (e.g. from 2.6 to 3.0) typically leads to a loss of progress. Therefore, unless the [release's page](https://github.com/axelboc/anki-ultimate-geography/releases) tells you otherwise, it is recommended to perform a clean import by following these steps:
+Major versions (e.g. `v3.0`) typically indicate that upgrading may lead to a significant loss of progress, or to the loss of [some of the customisations](#customising-the-deck) you may have made to the deck that would normally be preserved. This can occur, for instance, when [significant changes](https://github.com/axelboc/anki-ultimate-geography/releases/tag/v3.0) are made to the structure of the deck, or when [lots of cards](https://github.com/axelboc/anki-ultimate-geography/releases/tag/v4.0) are removed.
```
ohare93, line 148:
```suggestion
Major versions (e.g. `v3.0`) typically indicate that upgrading may lead to a significant loss of progress, or to the loss of [some of the customisations](#customising-the-deck) you may have made to the deck that would normally be preserved. This can occur, for instance, when [significant changes](https://github.com/axelboc/anki-ultimate-geography/releases/tag/v3.0) are made to the structure of the deck, or if our include criteria change which can result in [removal of some less relevant cards from the deck.](https://github.com/axelboc/anki-ultimate-geography/releases/tag/v4.0)
```

Read to me like card removal is a common thing that happens all the time! :sweat_smile:

README.md
```diff
@@ -149,3 +171,84 @@ Every note is tagged. The tags, which are listed below by category, are prefixed
...
+
+## Customising the deck
+
+Interested in making changes to the deck in Anki? Here's what you need to know.
+
+### Changes that are or can be preserved 💖
+
+- **Adding your own cards and media files**
+
+  Go wild. 🌊 Upgrading will not affect any new cards you create, or any new media files you add.
```
ohare93, line 183: User added media files will not be deleted from their Anki by ugrading, however their inclusion in any of the existing cards in the fields themselves can and probably will be overwritten.

README.md
```diff
@@ -149,3 +171,84 @@ Every note is tagged. The tags, which are listed below by category, are prefixed
...
+
+  If a duplicated deck is [created on import](#first-upgrade-after-apkg-import), your own cards will stay in their original deck.
+
+- **Moving cards into another deck**
+
+  Some people prefer to combine multiple decks into one to review them together.
+
+  By default, CrowdAnki will move all the cards back into the _Ultimate Geography_ deck on import. To prevent this behaviour and update the existing cards in place, follow the steps in the [_Getting started_](@getting-started) section but in the last step, make sure to tick the _Do Not Move Existing Cards_ checkbox in the _CrowdAnki Import Settings_ dialog box.
```
ohare93, line 191:
```suggestion
  By default, CrowdAnki will move all the cards back into the _Ultimate Geography_ deck on import. To prevent this behaviour make sure to tick the _Do Not Move Existing Cards_ checkbox in the _CrowdAnki Import Settings_ dialog box to update the existing cards in place, when following the steps in the [_Getting started_](@getting-started) section.
```

Give the pertinent info asap. Helps greatly in these long ReadMe files when people are scanning for the info the want.

README.md
```diff
@@ -149,3 +171,84 @@ Every note is tagged. The tags, which are listed below by category, are prefixed
...
+
+  - suspending or burying the card;
+  - creating a [filtered deck](#custom-study) that doesn't include the card;
+  - moving the card into an "icebox" deck and ticking the _Do Not Move Existing Cards_ the next time you upgrade, as explained in the previous section, under _Moving cards into another deck_.
+
+### Changes that prevent upgrading ⛔
+
+- **Adding a new field (population, currency, etc.)**
+
+  Adding a new field to the _Ultimate Geography_ note type will prevent you from upgrading the deck entirely. When you'll try to import it again, Anki will likely error and you'll risk losing your entire progress. Hopefully Anki, CrowdAnki, Brain Brew and UG will [one day](https://github.com/ohare93/brain-brew/issues/4#issuecomment-644975261) find a way to make this possible.
```
ohare93, line 250:
```suggestion
  Adding a new field to the _Ultimate Geography_ note type will prevent you from upgrading the deck entirely. When you'll try to import it again, Anki will likely error and you'll risk losing your entire progress. Hopefully Anki, CrowdAnki, Brain Brew, and UG will [one day](https://github.com/ohare93/brain-brew/issues/4#issuecomment-644975261) find a way to make this possible.
```

Oxford Comma! :wink: 

And god damn, linking people to the mega thread is just mean :rofl: :+1:

src/headers/default.yaml
```diff
@@ -6,7 +6,7 @@ extendRev: 50
 deck_configurations:
 - __type__: DeckConfig
   crowdanki_uuid: 43d37fac-9a65-11e8-9cc5-a0481cc15658
-  name: Default
+  name: Ultimate Geography
```
ohare93, line 9: Whoops, how was this still named Default? :sweat_smile:

README.md
```diff
@@ -18,15 +18,20 @@ The deck is available in **English**, **German**, **Spanish**, **French**, **Nor
...
-  - [Minor version](#minor-version)
+  - [First upgrade after APKG import](#first-upgrade-after-apkg-import)
+  - [Keeping deck customisations](#keeping-deck-customisations)
+  - [Levelling up from standard to extended](#levelling-up-from-standard-to-extended)
   - [Major version](#major-version)
 - [**Deck structure**](#deck-structure)
+- [**Customising the deck**](#customising-the-deck)
+  - [Changes that are or can be preserved](#changes-that-are-or-can-be-preserved-)
+  - [Changes that get reverted](#changes-that-get-reverted-)
+  - [Changes that prevent upgrading](#changes-that-prevent-upgrading-)
```
ohare93, line 30: Useless -s at the end of the link? I guess they are no harm.

ohare93 (commented):
Looks great, fantastic work :+1:

README.md
ohare93: And then if they are not references Anki will remove them on a Media check.

README.md
```diff
@@ -87,57 +92,74 @@ Anki's [search feature](https://docs.ankiweb.net/#/searching), coupled with the
...
+### Keeping deck customisations
 
-> For more information about versioning and the differences between minor and major, please refer to the [_Versioning_](CONTRIBUTING.md#versioning) section.
+If you've made, or are thinking of making, any changes to your _Ultimate Geography_ deck, check the [_Customising the deck_](#customising-the-deck) section below to see what will happen to them the next time you upgrade, and whether you need to take any steps to preserve them.
 
-### Minor version
+### Levelling up from standard to extended
 
-For a minor version upgrade (e.g. from 2.3 to 2.6), _do not_ import the APKG package in Anki as you may have done initially or you will lose your progress. Instead, proceed as described in the [_Other languages and versions_](#other-languages-and-versions) section, making sure to pick the correct archive (e.g. `Ultimate_Geography_v[...]_EN.zip` for the standard English deck).
+Importing an extended deck on top of a standard deck may bring unexpected results. You should consider learning the extended deck from scratch. To do so, either:
```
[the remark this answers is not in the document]
aplaice: If the users just click-through all the dialogues (just selecting OK), which many people will do (and I don't blame them — given the general preponderance of confirmations dialogues), then they will get unexpected results (they'll end up having a jumbled-up review schedule (with what are really new cards ("Country - Flag") presented as old cards, and vice-versa, which is arguably far worse than starting from scratch).

You're right, though, that we might wish to soften the language slightly.

Perhaps:

`may bring unexpected results` → `can be tricky and needs care`

and

`You should consider learning the extended deck from scratch` → `You may instead consider learning the extended deck from scratch`

?

README.md
aplaice: Perhaps:

```suggestion
  Go wild. 🌊 Upgrading will not affect any new cards you create, or any new media files you add to the newly added cards.
```

?

README.md
axelboc: I'm actually tempted to remove this example completely. After all, we'll never remove as many cards as the last time we did. The seas criteria will only remove a couple of cards, which I don't think will justify a major release.

README.md
axelboc: Good call, though I think "and update the existing cards in place" needs to stay put for the sentence to remain clear:

> To prevent this behaviour and update the existing cards in place, make sure to tick the _Do Not Move Existing Cards_ checkbox in the _CrowdAnki Import Settings_ dialog box, when following the steps in the [_Getting started_](@getting-started) section.

src/headers/default.yaml
axelboc: Yeah, makes it very tricky to distinguish it from Anki's _Default_ option group 😄

README.md
axelboc: I believe this is how GitHub identifies anchors to headings that end in `[space][emoji]`. I'll fix on `master` if it's not the case after merging.

README.md
axelboc: We may remove some of the lakes with seas in their name, but again, not a huge removal.

axelboc:
Latest changes in https://github.com/axelboc/anki-ultimate-geography/pull/403/commits/03ab5aefe756bf9123f01a72cf83454259eb8add

README.md
```diff
@@ -149,3 +171,84 @@ Every note is tagged. The tags, which are listed below by category, are prefixed
...
+
+  When importing the deck for the first time, CrowdAnki creates a new "option group" called _Ultimate Geography_ (unless it already exists). If you've made any changes to this group's options, they will be reverted on import.
+
+  To avoid losing your options, create your own option group and give it a distinctive name. After you upgrade, you'll just have to switch the deck back to this option group.
+
+- **Customising the templates and styles**
+
+  Any edits you make to the built-in templates of the _Ultimate Geography_ note type will be reverted on upgrade. Same goes with the CSS styles that are shared by all the templates.
+
+  One way to workaround this and preserve your changes is to  first duplicate the note type and switch all the cards to the new note type in the Anki browser. You can then make your template and style changes in the duplicate note type without affecting the original note type. The next time you upgrade, all the cards will switch back to the original note type, but you can easily switch them back to your customised note type in the Anki browser.
```
aplaice, line 207: Yes, this is far better (and more future-proof) than my approach!

Perhaps we should phrase it as "cloning" the note model, as that's how it's called in the interface as well as in the manual? Also, should we link the manual in case people don't know how to do it (I barely remembered and I've done it before...)?

```suggestion
  One way to workaround this and preserve your changes is to first [clone the note type](https://docs.ankiweb.net/#/editing?id=adding-a-note-type) and switch all the cards to the new note type in the Anki browser. You can then make your template and style changes in the cloned note type without affecting the original note type. The next time you upgrade, all the cards will switch back to the original note type, but you can easily switch them back to your customised note type in the Anki browser.
```

(It's possible that I'm missing an easier way of duplicating note models...)

<hr/>

I've tested it, and it works as desired and expected.

However, I'm slightly worried, as the `crowdanki_uuid` of the "cloned" note model is the same as that of the original one. (I've checked by exporting the deck with the cloned note model and by inspecting `collection.anki2` with sqlite.) (This is analogous to https://github.com/Stvad/CrowdAnki/issues/118, just for note models...) Hence, in principle, `CrowdAnki` could "update" the new (cloned) note model, rather than the original.

Fortunately, it seems (looking at [the UUID code](https://github.com/Stvad/CrowdAnki/blob/a583fa8ceaa4932cf00630476e357ef2e420233e/crowd_anki/utils/uuid.py#L34)) that `CrowdAnki` finds the first matching note_model (the first one having a matching UUID), and assuming that Anki keeps its note_models in chronological order, this will be the original note_model (as we desire), so there _shouldn't_ be problems (but I'm still not 100% comfortable...).

README.md
aplaice: We might also end up removing some ("geographical" rather than "political") islands.

axelboc: Ah right. All in all, I still don't think it will amount to "losing a significant amount of progress", especially considering that removed cards are not actually removed from people's decks on import.

I think I'll remove the quote that points to the `CONTRIBUTING` guide's _Versioning_ section. I don't think this section provides any more information than is already in the `README`.

README.md
```diff
@@ -149,3 +171,84 @@ Every note is tagged. The tags, which are listed below by category, are prefixed
...
+
+  > Note that you may have to update your custom templates and styles afterwards to stay in line with the built-in templates.
+
+- **Adding or removing a template**
+
+  We're getting into dangerous territory here, but you should be able to make this work by duplicating the default note type as explained in the previous point, _Customising the templates and their styles_.
+
+  The difference is that, when you import the deck and/or when you switch back the cards to your customised note type, you may see a dialog box asking you to match the templates of one note type to the templates of the other note type. (This same dialog box appears when importing the extended deck [on top of the standard deck](#levelling-up-from-standard-to-extended).)
+
+  As long as you map the templates that Anki asks you to match correctly, you should be fine. However, we strongly recommend that you try this out in a separate Anki profile first.
```
aplaice, line 217: tldr; Do not recommend :scream:

I might be doing something wrong, but I don't see how upgrading a deck with an added card template can be done without loss of progress (without manually fiddling with the `deck.json` beforehand (my workaround of first exporting your current deck and then re-importing doesn't really work (or perhaps it works, while crashing Anki??)!)).

With the UUID cloning issue, the note model is (for me) updated "in-place" without any dialogue, unlike in the Standard → Extended case (or unlike if the original model had had an extra card added), since here 

1. the UUIDs of the two note models are the same
2. the original note model is not changed (no cards/fields added/removed)

Hence, all the notes are silently moved back to the "original" note model and silently have one card template removed, (without the corresponding "extra"/now attached cards removed). As a result, when I open the browser, I get a "list index out of range" error (since there are cards attached to the "original" note model which have an index higher than the number of cards/note...) and can't do anything. If I try to import an export of the deck with the cloned note model, then I get a key error (again because there are cards that have no right to exist...) (though after killing Anki and restarting, it seems that all cards including the ones from the added templates are restored, again, with progress). Alternatively, (instead of trying to import an export of the deck with the cloned note) I can check the database, but that deletes the extra cards.

If the UUID cloning issue were not present, (i.e. the cloned deck had its own UUID, rather copying that of the original) then you'd be presented with the standard note model migration dialogue (like for Standard → Extended) and would have to force mapping the added cards (those corresponding to the added template) to "Nothing"...

<hr/>

Removing templates should be fine.

README.md
axelboc: Hmm, that's disappointing. That being said, the worst that can happen is that CrowdAnki updates the cloned note type, in which case the user would just lose their customisation. Maybe I can just add a warning that this may not work in the future or something and to recommend backing up the deck before upgrading.

README.md
```diff
@@ -149,3 +171,84 @@ Every note is tagged. The tags, which are listed below by category, are prefixed
...
+
+  Upgrading will bring the card back. If you don't want to learn a card at all or if you want to delay learning it, consider:
+
+  - suspending or burying the card;
+  - creating a [filtered deck](#custom-study) that doesn't include the card;
+  - moving the card into an "icebox" deck and ticking the _Do Not Move Existing Cards_ the next time you upgrade, as explained in the previous section, under _Moving cards into another deck_.
+
+### Changes that prevent upgrading ⛔
+
+- **Adding a new field (population, currency, etc.)**
```
aplaice, line 248: Ugh, yeah, you're right. This (the fact that you get a `list index out of range` error rather than having the extra field successfully mapped to Nothing) is a `CrowdAnki` bug. :(

README.md
aplaice: > this may not work in the future

Hopefully, I'll get around to fixing https://github.com/Stvad/CrowdAnki/issues/118 before sufficient things change in the Anki internals that this stops working.

> recommend backing up the deck before upgrading

FWIW `CrowdAnki` does a backup before every import, but doing one manually won't hurt.

aplaice (approved):
Summing up, this looks really useful! :)

README.md
axelboc: You're right, I get the error as well. I think when I tested, I must have added one template and removed another, which led to the same number of templates and may explain why I didn't see the error.

axelboc: Removing templates by cloning the note type ends up being more complicated than just removing the template again after the upgrade. So I'll move it into the next section.

axelboc:
The latest batch of changes: https://github.com/axelboc/anki-ultimate-geography/pull/403/commits/b5e3fcd1548f3e9ecffa3c679140b6b81fdc809f

PR #409, opened 2021-04-05
Changes Capital of Palestine and adds hints to Israel and Palestine

### LINKED ISSUE #389: Capital of Israel is disputed
Hi,

Israel's capital is listed as Jerusalem, with no further note. Wikipedia describes it as disputed: “Recognition by other UN member states: Australia (West Jerusalem),[1] Russia (West Jerusalem),[2] the Czech Republic (West Jerusalem),[3] Honduras,[4] Guatemala,[5] Nauru,[6] and the United States.[7] In September 2020 it was reported that Serbia would be moving its embassy from Tel Aviv to Jerusalem.[8][9]” with link to: https://en.wikipedia.org/wiki/Status_of_Jerusalem

Suggested note (all of this is from the given Wikipedia article, with some parts removed that do not change the meaning, for brevity): “Disputed, with both Israel and Palestine claiming Jerusalem as capital. The majority of UN member states have favored locating their embassies in Tel Aviv.”

jamesnelmore:
Closes #389 with most current agreed to solution


jamesnelmore:
Closes #389 with most current agreed to solution

aplaice (approved):
Thanks very much for preparing a pull request!

I still have doubts regarding the capital hint, but I don't have better suggestions, so this version makes sense.

Given that we seem to be following British English spelling conventions, I'd suggest switching to "administrative cent*re*", but it's a minor issue.

I'd also suggest keeping the Norwegian capital info for Palestine, since it seems to already partially say what we want it to say.

jamesnelmore:
I added back the Norwegian translation and changed the hint to British English. I was uncertain about keeping the translation because I do not speak Norwegian, however I used Google translate to confirm that it does say what it appears to say.

aplaice:
Thanks very much!

I think that the failing check is because the Norwegian country info contains a comma, but the field isn't quoted. 

In brief, CSV uses `,` as a delimiter between fields. Often you want a comma within a field. In these cases, in order to be able to distinguish a field-delimiting comma and a comma within a field, you quote the whole field with `"`.

Hence, you'd need `"Jerusalem er proklamert hovedstad, mens Ramallah er administrativ hovedstad."`. Otherwise, the integrity thinks that there's an extra column (due to the extra unquoted comma) and is confused.

(Yes, it's a bit annoying! :/)

axelboc (changes requested):
Thanks for the PR!

I think we should add the missing translations as best as we can before merging as per my comment here: https://github.com/anki-geo/ultimate-geography/pull/388#pullrequestreview-632919783. Machine translations will do.

Happy to do it, probably tomorrow, unless someone else has some time to spare before then.

jamesnelmore:
The fields have been translated.

axelboc (approved):
Fantastic, thanks for that 🎉

PR #414, opened 2021-04-10
Remove requirements.txt and use Pipenv in CI

### LINKED ISSUE #413: Remove requirements.txt
There's a Dependabot alert caused by an outdated package version in `requirements.txt`. (It's up to date in `Pipenv.lock`, though.)

I've asked about this file before: https://github.com/anki-geo/ultimate-geography/pull/364/files#r550776335. Reading a bit more about dependency management in Python, it seems that `requirements.txt` is on the way out. Python even recommends `Pipenv` in [its own docs](https://packaging.python.org/tutorials/managing-dependencies/#managing-dependencies)

It seems counter-productive to have to keep `requirements.txt` up to date with `Pipenv.lock`, especially since GitHub is capable of [parsing the latter](https://github.com/anki-geo/ultimate-geography/network/dependencies) without problem:

image

Also, I think having the same set-up code running in the CI as the one recommended in `CONTRIBUTING.md` would make more sense. I don't believe installing `Pipenv` in the CI would impact the build time, thanks to GitHub Actions' caching system.

axelboc:
Fix #413 

I'm using this GitHub Action to simplify the CI workflow: https://github.com/marketplace/actions/pipenv-for-github-actions

### DIFF
--- a/.github/workflows/integrity-check.yml
+++ b/.github/workflows/integrity-check.yml
@@ -9,17 +9,20 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
+
     - name: Set up Python 3.8
       uses: actions/setup-python@v1
       with:
         python-version: 3.8
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install -r requirements.txt
+
+    - name: Install dependecies
+      uses: VaultVulp/action-pipenv@v2.0.1
+      with:
+        command: install
 
     - name: Validate
-      run: |
-        brain_brew recipes/source_to_anki.yaml || (echo "::error file=source_to_anki,line=0,col=0::Build failed" && exit 1)
+      uses: VaultVulp/action-pipenv@v2.0.1
+      with:
+        command: run build
 
       # TODO: Validate the csv file structure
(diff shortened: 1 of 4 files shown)


ohare93:
As I said here https://github.com/anki-geo/ultimate-geography/issues/413#issuecomment-817149707 I think requirements.txt is currently used by Github.

1. To give Dependency warnings (See "requirements.txt" under PyYaml)
image

2. To show the dependencies and dependents of a repo, which are also visible on the front page of the repo (though I cannot see them on UG right now for some reason)
image

These are both useful functions.

If Github is indeed able to do these without the dependencies.txt file, then I am all in for removing it. But I cannot find a concrete answer for that when googling around :face_with_head_bandage:

axelboc:
The screenshot I took in the issue is from the "Dependencies" tab. So yes, Dependabot can find dependency issues in `Pipenv.lock` and GitHub shows the dependencies listed in `Pipenv.lock`. From this, I assume that GitHub will still show this package as a dependent of Brain Brew once we remove `requirements.txt`. If not, we can revert.

ohare93 (approved):

axelboc:
Got it to pass 😤 I kept on making typos.

I had to add a script to the `Pipfile` that the CI action could run. I think it's a convenient thing to have anyway: `pipenv run build`. I mention this new script in the contributing guide.

ohare93 (approved):

Issue #416, opened 2021-04-11
Capital inconsistencies between languages


axelboc:
Looking at #409, I noticed that the proclaimed capital of [Palestine](https://en.wikipedia.org/wiki/State_of_Palestine) is actually inconsistent on Wikipedia across languages:

- [Jerusalem](https://en.wikipedia.org/wiki/Jerusalem) on EN, NB, CS, RU (both listed), NL
- [East Jerusalem](https://en.wikipedia.org/wiki/East_Jerusalem) on DE, ES, FR, RU (both listed), SV

It's an even split! 😂 I decided not to fix it right away, as I thought it required further discussion... so here I am.

It actually reminded me of other capital inconsistencies we've discussed in the past, like Kiribati #166... So for fun, I decided to do a quick review of all the capital inconsistencies currently in the deck:

- **Montenegro**: `ES` has "Podgorica, Cetiña" but all other languages have only Podgorica.
- **Palau**: `FR` has "Melekeok" but all other languages have Ngerulmud.
- **European Union**: `NB` has "Brussel, Strasbourg, Luxembourg" but all other languages have only Brussel.
- **Sri Lanka**: `NB` has "Colombo" but all other languages have Sri Jayewardanapura Kotte. Norwegian Wikipedia now lists both: "Sri Jayewardanapura Kotte, Colombo"; it's better but it's still different.
- **Tuvalu**: `NB` has "Vaiaku/Funafuti" but all other languages have only Funafuti. Fortunately, this has been fixed on Norwegian Wikipedia.
- **Kazakhstan**: `NB` has "Astana" but all other languages have Nur Sultan. Fortunately, this has been fixed on Norwegian Wikipedia.
- **Kiribati**: `NB` has "Bairiki (på Tarawa øy)" but all other languages have Tarawa.  Wikipedia has switched to South Tarawa _almost_ across the board, including in Norwegian, but the capital is written as "Tarawa (Jižní Tarawa)" - i.e. _Tarawa (South Tarawa)_ - on Czech Wikipedia.

---

Looking at all of this, there are a number of incorrect or outdated capitals that clearly need to be fixed (Palestine, Sri Lanka, Tuvalu, Kazakhstan and Kiribati).

Assuming we do (I'll open a PR), we'll be be left with the following inconsistencies:

1. one where half of the languages have a different capital than the others: **Palestine**;
[...]

PR #417, opened 2021-04-11
Fix outdated capitals in Norwegian

axelboc:
See #416.

### DIFF
--- a/src/data/capital.csv
+++ b/src/data/capital.csv
@@ -54,7 +54,7 @@ Ukraine,Kyiv,Kiew,Kiev,Kiev,Kiev,Kyjev,Киев,Kiev,Kiev
 Vatican City,Vatican City,Vatikanstadt,Ciudad del Vaticano,Vatican,Vatikanstaten,Vatikán,Ватикан,Vaticaanstad,Vatikanstaden
 Armenia,Yerevan,Jerewan,Ereván,Erevan,Jerevan,Jerevan,Ереван,Jerevan,Jerevan
 Cyprus,Nicosia,Nikosia,Nicosia,Nicosie,Nikosia,Nikósie,Никосия,Nicosia,Nicosia
-Kazakhstan,Nur-Sultan,Nur-Sultan,Nur-sultán,Noursoultan,Astana,Nur-Sultan,Нур-Султан,Nur-Sultan,Nur-Sultan
+Kazakhstan,Nur-Sultan,Nur-Sultan,Nur-sultán,Noursoultan,Nur-Sultan,Nur-Sultan,Нур-Султан,Nur-Sultan,Nur-Sultan
 Egypt,Cairo,Kairo,El Cairo,Le Caire,Kairo,Káhira,Каир,Caïro,Kairo
 Algeria,Algiers,Algier,Argel,Alger,Alger,Alžír,Алжир,Algiers,Alger
 Angola,Luanda,Luanda,Luanda,Luanda,Luanda,Luanda,Луанда,Luanda,Luanda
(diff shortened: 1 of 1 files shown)


aplaice:
We'd also need to update our [Norwegian contributing guidelines](https://github.com/anki-geo/ultimate-geography/blob/master/CONTRIBUTING.md#norwegian-bokm%C3%A5l), since there Wikipedia is only a source of third priority. The current Norwegian capitals in the deck are indeed consistent with the first-priority source.

I 100% agree with Bucharest no longer being the capital of Russia. :)

aplaice:
Given that:

1. the [highest priority source](https://www.regjeringen.no/no/dokumenter/statsnavn-hovedsteder-og-nasjonaldager/id87863/) was last updated in 2013, 
2. the [second highest priority source](https://snl.no/) at least sort-of agrees with Wikipedia in all three cases:
    For [Kazakhstan](https://snl.no/Kasakhstan) it has a different spelling (Nursultan in snl vs. Nur-Sultan in Wikipedia), but no mention of Astana.
    For [Sri Lanka](https://snl.no/Sri_Lanka)  it has "Colombo og Sri Jayewardenepura Kotte", which is slightly weird.
    For [Tuvalu](https://snl.no/Tuvalu) it agrees completely.

it might make sense to follow Wikipedia (or SNL?). I'd consult with @Vages first, though.

axelboc:
Yeah good point, let's wait on @Vages' input.

Some of the inconsistencies are caused by the presence of a paren block that gives an alternate name or spelling. I've actually just spotted another one: `Helsingfors (Helsinki)`. In such case, if the source order remains the same, then we could at least apply the rule of moving the alternate name/spelling to the _Capital info_ field, along with additional information (if there isn't already).

That being said, I feel like the first two sources will never be as up to date as Wikipedia. `Nur-Sultan is a good example. I think it wasn't clear at first whether it was going to be spelt with a hyphen or not, but then it got clarified by official Kazakh institutions. So I'd be in favour of putting Wikipedia first and of using the other two sources as back-ups for any missing/unclear information.

To support my argument, I'll quote the current capital info of Kiribati that Vages wrote, which I've translated back into English:

> Many consider Bairiki as the capital of Kiribati (also the Ministry of Foreign Affairs in Norway) because both the parliament and the president once had a seat there. Since the year 2000, the parliament and ministries have been spread across South Tarawa.

This is basically saying that the Ministry of Foreign Affairs in Norway is wrong and that Wikipedia is right... So we should probably make it official. 😄

aplaice (approved):

aplaice:
> In such case, if the source order remains the same, then we could at least apply the rule of moving the alternate name/spelling to the Capital info field, along with additional information (if there isn't already).

Yeah, that makes sense.

<hr/>

> That being said, I feel like the first two sources will never be as up to date as Wikipedia.

FWIW the SNL article on Kazakhstan was last updated on 14 January 2021, so it's not that old, but indeed it won't ever be as up-to-date as Wikipedia. OTOH capital and country info don't change _that_ often, so frequent updates mostly aren't that essential.

<hr/>

Ugh, I can't explicitly add Vages as a reviewer.

axelboc:
I've added all the users who contributed translations as members of this repo with Read access. Hopefully, this means we'll be able to add them as reviewers (once they've accepted the invitations).

Vages:
Your arguments for prioritizing SNL over the Ministry of foreign affairs are sound. I'll summarize as follows:

- Ministry of foreign affairs may have diplomatic reasons for lagging the _de facto_ changes. Thay also don't seem to update things that often.
- Store norske leksikon (SNL) have professional editors, usually academics without a direct bond to the government. In my opinion the most authorititative source.
- Wikipedia is mostly correct and usually the first source to be updated, and quickly corrected. «Wrong, but not for long»

I'll update the contributing guidelines to reflect this.

axelboc:
Sorry, haven't had time to come back to this. Given @Vages feedback, and our decision to use English Wikipedia as our official source for factual information https://github.com/anki-geo/ultimate-geography/issues/416#issuecomment-821864712, I'm going to close this PR and deal with all the factual inconsistencies at once.

PR #418, opened 2021-04-11
Change capital of Kiribati to South Tarawa

axelboc:
As @ohare93 had mentioned in https://github.com/anki-geo/ultimate-geography/issues/166#issuecomment-632590049, Wikipedia changed back to South Tarawa a while back and the change hasn't been reverted since. Also, all languages seem to agree. See #416 for more details.


aplaice (approved):
FWIW I disagree with Wikipedia — as mentioned in [the other thread](https://github.com/anki-geo/ultimate-geography/issues/166#issuecomment-632628269) all the citations for "South Tarawa" give the capital as Tarawa, but given that it's been 9 months since [the last (?) switch to South Tarawa](https://en.wikipedia.org/w/index.php?title=Kiribati&diff=965942280&oldid=965887785), it probably makes sense to go back to "South Tarawa"...

axelboc:
Oh well, seems that someone (probably from this deck) just reverted English Wikipedia back to Tarawa. I've fixed a couple of remaining references to South Tarawa as being the capital, and I've also made the change from `Tarawa-Sud` to `Tarawa` on French Wikipedia.

We're still out of sync with Wikipedia in most languages, but I don't think it's worth getting back in sync for the time being...

PR #422, opened 2021-04-19
Update Norwegian contribution guidelines

Vages:
In reaction to #417


aplaice (approved):
Makes sense — you have a better idea of how good or bad Norwegian Wikipedia is, and what other Norwegian sources are reputable.

[We'll likely be switching to using English Wikipedia as authoritative for factual information, for all languages](https://github.com/anki-geo/ultimate-geography/issues/416#issuecomment-821864712), while still using the respective language sources for spelling and alternative names. For instance, if English Wikipedia claims that New-Town is the capital of Examplestan, while SNL claims that it's Gammelstad but states that the Norwegian name for New-Town is Nystad, sometimes also known as Alternativstad, then in the Norwegian deck, we'll have Nystad as the capital of Examplestan, mention Alternativstad as an alternative name in the *Capital info* and probably, (if the change from Old-Town to New-Town was recent) mention Gammelstad as a previous capital.

(Apologies if I've butchered Norwegian in my made-up names...)

The idea is that the actual identity of the capital (or the best guess of what it is) doesn't depend on language — just the name(s) do.

Hopefully you don't have any objections, but if you do, please say so, and we'll try to find a solution.

PR #424, opened 2021-05-08
correct german country info of Hongkong

x51mon:
grammer was incorrect
added full country name of china, same like in the note for Macau


aplaice (approved):
Oops and thanks!

In case @axelboc prefers a shorter name (in line with what we have in other languages, both for Macau and Hong Kong), would:

```
Sonderverwaltungszone Chinas.
```

also be correct?

x51mon:
> ```
> Sonderverwaltungszone Chinas.
> ```

yes that'd also be correct

PR #427, opened 2021-05-09
Add a link to a blank world map to desc.html

whnr:
I found having a blank world map available extremely helpful for pointing at a physical map when being asked where a card is located.
So I linked to one resource on wikimedia that is similar to the map in the deck

### DIFF
--- a/src/headers/desc.html
+++ b/src/headers/desc.html
@@ -12,4 +12,6 @@ The deck is available in <a href="https://github.com/axelboc/anki-ultimate-geogr
 
 You can use Anki's <a href="https://github.com/axelboc/anki-ultimate-geography#custom-study">filtered deck feature</a> to focus your study on a subset of the deck, such as sovereign states, a single note template (e.g. map to country), or a specific continent (e.g. Europe).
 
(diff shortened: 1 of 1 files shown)


axelboc (commented):
That's when learning the extended deck's _Country - Map_ cards, right? If so, then perhaps the deck's description is not the right place, as it is the same on the normal deck. How about moving this to the README, in the _Features_ section, where the extended deck is mentioned?

whnr:
@axelboc, the reason why I put it in this place is that it provides a quick way to get to the map that is integrated into anki. I have been a happy user, but might also be the only person that likes it there.

I don’t know how fancy anki is with css styles or how the build process for the extended and normal deck work, but there could be two options:

1. Two different desc.html files for each type of deck (the current has redundant info about the extended deck when you are already using it)
2. Just show or hide some html elements with a `.normalDeck {display: none}` style

axelboc:
I'm going to close this for now, sorry. 

Hopefully this is a use case that our deck manager, BrainBrew, can support in the future. 🤞

PR #433, opened 2021-06-27
Fixed some Dutch translation errors

lennardnl:
The Dutch translation of the countries and capitals contained some errors, which I have corrected.


axelboc (approved):
Nice, well spotted! 🎣

aplaice (approved):
Thanks very much for catching these!

(I've double-checked that the updated capital cities now indeed have the correct spellings, as they are on Dutch Wikipedia.)

PR #434, opened 2021-07-01
Change the map of the Baltic Sea to follow the IHO 2002 draft

aplaice:
See: http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=34

It now includes the Kattegat and the Danish Straits.

The current (old) map is just _wrong_ — it follows neither the IHO 1953 publication, nor the IHO 2002 draft, or any other logical limit of the Baltic Sea.  It excludes the Kattegat and the Danish Straits, like IHO 1953, but unlike IHO 2002, but also excludes some of the sea between the Danish islands and Germany, which is part of the Baltic Sea under even the most restrictive limits.

The reason for the old, incorrect map was because these were the best boundaries that could be obtained, by combining various polygons, from Natural Earth Data.  (Natural Earth Data has a polygon for "Internal Denmark Waters", which combines some of the Danish straits and part of the sea between the straits and Germany, while the polygon for the Kattegat includes part of the Skagerrak.)

[Pangaea](https://doi.pangaea.de/10.1594/PANGAEA.777975) (under CC-BY) is the data source for the new map.  I hadn't used it previously since I was aiming to stay as close as possible to IHO 1953 rather than to IHO 2002, but we've now explicitly switched to IHO 2002.  (Note that the source claims to follow IHO 1953, but in the case of the Baltic Sea it's not.)

(I've extremely belatedly realised that GitHub has a rich diff image viewer, so I'm not including the images inline...)


aplaice:
I've also asked again (after 1.5 years) for guidance regarding pushing all the SVG sea maps to Wikimedia [here](https://en.wikipedia.org/wiki/Wikipedia_talk:WikiProject_Maps#SVG_Locator_maps_of_seas), so hopefully I'll get some reply and will finally "upstream" all these maps!

(If I don't get a reply, I'll probably push them anyway in a couple of weeks — the total ≲ 0.5 GB of the SVGs is a lot to push without consultation, but not absurdly so.)

axelboc:
Awesome! There's a tiny bit of sea that is removed in the new map:

image image

I can't tell whether it should or not at this level of detail:

image

axelboc (approved):
Anyways, it's so tiny, it doesn't matter either way. Feel free to merge.

aplaice:
> There's a tiny bit of sea that is removed in the new map:

Oops! Thanks for noticing!

That's just a case of the creators of the shapefile and me being sloppy (them not including that area/me for not noticing). I don't think it's any special subdivision of the Baltic Sea, so there's no reason to exclude it. It's really easy to fix!

I've also now realised that I changed the state of the [Szczecin Lagoon](https://en.wikipedia.org/wiki/Szczecin_Lagoon). If we are to follow IHO 2002 as closely as possible, it should be excluded! The  [Curonian Lagoon](https://en.wikipedia.org/wiki/Curonian_Lagoon) should also be excluded, by the same argument (it's included in both the old and new maps).  It should be relatively straightforward to exclude both.

(English Wikipedia isn't 100% clear about whether the Lagoons are part of the Baltic Sea, though it seems to lean towards not. For comparison, though, German Wikipedia seems to think that they are.)

aplaice:
Hopefully I've fixed everything, but please nitpick if anything is still wrong!

axelboc:
Looks good to me! 💯 

Can't believe even this Pangaea shapefile has inaccuracies! They really don't want to make our life easy, do they!? 😂

PR #440, opened 2021-07-05
Remove EU capital and capital info

### LINKED ISSUE #438: Review capital(s) of the European Union
Currently, the Norwegian deck lists Brussels, Strasbourg and Luxembourg together as capitals of the EU, while the other decks give only Brussels. I was about to remove the two extra capitals in Norwegian as per https://github.com/anki-geo/ultimate-geography/issues/416#issuecomment-821864712 when I noticed that the infobox on English Wikipedia no longer mentions a capital at all! It just lists the various institutions and their locations in a section called "Institutional seats".

Should we then reconsider the decision we made in #213? The way I see it, we can either:

1. remove the capital from the _European Union_ note completely (to leave only the flag and the map); or
2. list all four "institutional seats" cities (including Frankfurt, the seat of the Central Bank).

Given that there are no qualifiers such as "executive" or "legislative" in the infobox, I feel option 2 would be messy. We'd have to least all the institutions in the _Capital info_ field... Going with option 1 would have the added benefit of removing a couple of hints.

Personally, I'd rather go with option 1, especially since I've never really seen the EU as having a capital in the first place. What do you all think?

axelboc:
Fix #438


aplaice (approved):
We can also remove a third of `capital_hint.csv`! :)

axelboc:
Oh right, I forgot about it 😄

PR #441, opened 2021-07-06
Disambiguate capital of Northern Cyprus in Swedish

### LINKED ISSUE #437: Review capital of Northern Cyprus in Swedish
The Swedish deck has _Nicosia_ as the capital of _Northern Cyprus_ instead of _North Nicosia_ like the other decks. I think this is a case where the new guideline of following English Wikipedia (cf. https://github.com/anki-geo/ultimate-geography/issues/416#issuecomment-821864712) justifies switching to _North Nicosia_ in Swedish.

The problem is that Swedish Wikipedia does not have an article for _North Nicosia_; only for [Nicosia](https://sv.wikipedia.org/wiki/Nicosia), so I'm not sure what to do.

---

As a side note, there's no capital hint to help distinguish the two instances of _Nicosia_ in the Swedish deck at the moment. So if we keep _Nicosia_ we should at least add a hint.

aplaice:
Thanks @emiham (and thanks @axelboc for noticing this in the first place)!

Fix #437.

I'm not touching the Dutch version of North Nicosia ("North Nicosia" vs. "Noord-Nicosia" or something similar), for the moment!


axelboc (approved):
Makes sense, thank you both.

PR #444, opened 2021-08-07
Russian translation for the capital of Maldives

zserghei:


axelboc (approved):
✔️ https://ru.wikipedia.org/wiki/%D0%9C%D0%B0%D0%BB%D0%B5

PR #445, opened 2021-08-07
Upstream most sea maps

aplaice:
### Upstream most sea maps

The license has changed from CC BY-SA 3.0 to CC BY-SA 4.0, because while uploading to Wikimedia I only realised half-way the upload process that the default license was CC BY-SA 4.0 not 3.0, and switching to 4.0 for only half the maps would have been worse than switching to 4.0 for all of them.

The Tasman Sea is sort-of upstreamed, but I've realised that in this case, I had previously been accidentally doing something more correctly than now:

The IHO definitions (and the shapefiles) contain "straight lines".  According to IHO 1953 these are supposed to be rhumb lines (geodesics/parts of great circles).

My map generation pipeline used to do sort-of the right thing (convert long line segments into approximate geodesics), under some circumstances, but fail catastrophically under others (not interpolate some long lines at all, resulting in, for instance, completely wrong Oceanic boundaries).  From what I remember, the relevant `densify` options didn't help, in these cases, so I started "densifying" long lines with a custom script.  This fixed the catastrophic failures, but because I was densifying incorrectly (assuming that the lines were in flat Euclidean space...), apparently also made some of the boundaries slightly worse.  Visually comparing all images, the deterioration seems to only have occurred in the case of the Tasman Sea.

Hence, I don't want to set the upstream of our Tasman Sea map to the new, slightly worse SVG.  Instead, I'll (hopefully, at some point) fix the densification script (it shouldn't be hard, just slightly fiddly), update the version uploaded to Wikimedia and then update sources.csv.

### Include ths Strait of Tartary in the Sea of Okhotsk

I remember previously deliberately excluding it, but I don't remember why, and since both IHO 1953 and IHO 2002 include it, I'm following them.

### Improve Gulf of Carpentaria map

It's now in line with IHO 2002. (1953 didn't have it.)

### Improve Gulf of California map (southern limit)

Now its limits are in line with both IHO 1953 and 2002.

### Improve the map of the East Siberian Sea slightly

It's now more-or-less conformant with IHO 1953.  It still differs from IHO 2002 considerably, but slightly less so. (Edit: I've realised that the geodesic issue also affected the East Siberian Sea, but in this case, I think that the new map is still a mild improvement, on net.)


axelboc (approved):
Impressive work, well done 👏👏👏

aplaice:
Thanks!

PR #453, opened 2021-08-11
Change name of capital as adapted to Spanish

josealberto4444:
It seems that the name has been adapted to Spanish recently.


aplaice (approved):
Thanks!

axelboc (approved):

PR #457, opened 2021-08-18
Improved Russian translation

zserghei:


aplaice (approved):
Thanks so much @zserghei! 

To the extent that I could double-check on Wikipedia, this looks great!

I've resolved the spurious "conflict" with master.

I'm not sure if you've finished — if not, please just open another PR! (It'd be great to have the fixes you've already made in the upcoming 5.0 release, so I merged now rather than later.)

<hr/>

I squashed to avoid having the multiple merge into master commits (due to us having updated `anki-geo:master` while you were working on your branch) and to avoid having the `src/данные/country.csv` <-> `src/data/country.csv` renaming in the history.

I hope squashing was OK! (Thanks again!)

PR #458, opened 2021-08-19
Switch python version to 3.7

aplaice:
(This is what's specified in our Pipfile and BrainBrew's.)

Hopefully, this will fix our integrity checks! (or not...)

(I'm not sure why/how they had been working previously — I'd guess that it was because Python 3.7 was already available in the "base image", and now the base image has been upgraded.)

(Edit: the python version should be completely irrelevant since `action-pipenv` runs in a docker container anyway...)

### DIFF
--- a/.github/workflows/integrity-check.yml
+++ b/.github/workflows/integrity-check.yml
@@ -8,10 +8,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
 
     - name: Set up Python 3.7
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: 3.7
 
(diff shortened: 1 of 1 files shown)


axelboc:
Thanks, I was scratching my head on those CI failures. I see you've switched to draft -- is it because it's failing? I think perhaps it's because the changes to the workflow won't be picked up until the PR is merged. Feel free to merge whenever if that's the case 😉

aplaice:
I've switched to draft, as I feel like I'm barking up the wrong tree. `VaultVulp/action-pipenv` is apparently running `pipenv` in a docker container, so the version of python we set up in the previous step should be irrelevant.

The changes do seem to be picked up by GitHub's action runner — Python 3.7 is being set up, it's just not being picked up later (I'd guess because the next steps are within docker, anyway.)

aplaice:
I've switched to running pipenv directly ourselves (without an external action) — it might be slower (no pre-compiled docker image with pipenv already installed), but it seems to work.

If we really want, we could write a custom GitHub action for BrainBrew, but that's less urgent. :)

axelboc (approved):
Makes sense, thanks for getting this fixed! 👍

PR #460, opened 2021-08-29
Add commas delimiting empty fields for other languages

aplaice:
Otherwise all the language decks CS-SV (Czech, Dutch, Russian and Swedish) were broken!

(The issue was that due to the missing empty fields, BrainBrew generated the Celebes Sea note as:

```
["Name of Celebes Sea in XX", None, "", ...]
```

instead of as:

```
["Name of Celebes Sea in XX", "", "", ...]
```

and the `None` confused Anki.


axelboc (approved):
Good catch 🎣

PR #465, opened 2021-09-06
Improved Russian translation

zserghei:


aplaice (approved):
Thanks very much for the further fixes!

(In case somebody else is confused, like I was, briefly: the South Tarawa change is correct — as per #166, we'll be changing all the versions to South Tarawa (in the respective languages!).)

axelboc (approved):

axelboc:
Thanks @zserghei! I'm merging this batch of changes now, as I'm going to release v5.0. Any further changes will go into the next release.

PR #466, opened 2021-09-07
Improve map of Bali

### LINKED ISSUE #425: Map of Bali is missing the small world map
The map of Bali is missing the small world map which indicates the position of the map section.
It makes it easier to guess the card based on that and a bit harder to locate the card section.

josealberto4444:
Fixes #425 

I edited the sources.csv file, but I'm not sure if I had to do something more.

I also added the information for the maps I modified previously (Micronesia and Polynesia).


aplaice (approved):
It's great! Thanks! (Also, thanks for updating `sources.csv`.)

> I'm not sure if I had to do something more.

I don't think so.

PR #467, opened 2021-09-07
Disambiguate Sint[ -]Maarten in Dutch

### LINKED ISSUE #464: Distinguish Sint Maarten from Sint-Maarten in Dutch
In Dutch, the names of Sint Maarten (constituent country of the Kingdom of the Netherlands) and Saint-Martin (French overseas collectivity) are [Sint Maarten](https://nl.wikipedia.org/wiki/Sint_Maarten_(land)) and [Sint-Maarten](https://nl.wikipedia.org/wiki/Sint-Maarten_(Franse_Antillen)) respectively.

This is obviously not identical but still confusingly similar.

I'd suggest following the approach we've used to distinguish the two in languages where they're identical (e.g. Saint-Martin in French) and adding a disambiguating parenthetical.

For instance:

```
Sint Maarten (Koninkrijk der Nederlanden)
Sint-Maarten (Frankrijk)
```
or closely clinging to Wikipedia:

```
Sint Maarten (land)
Sint-Maarten (Franse Antillen)
```

<hr/>

If somebody wants to train themself to distinguish the two by the lack of the hyphen (tbh I can imagine a school teacher requiring students to do that...), then they can remove the parenthetical.

(Noticed while reviewing the Russian fixes.)

<hr/>

@WouterVdS If/when you have time (this isn't urgent! :)):

Do you think that this makes sense, and if so, would either of the above pairs be suitable?

aplaice:
Fix #464.

(Probably not for v4.2/v5.0.)

<hr/>

(I'm not sure if we actually want this change, but I've updated the branch just in case anyway.)


axelboc (approved):
Forgot about this one, thanks, looks good to merge 👍

PR #468, opened 2021-09-07
Upstream Tasman Sea map

aplaice:
The interpolation of geodesics has been corrected, so the upstream map is now a valid source of the current map.

(There are some, very slight changes so I've also updated the map in the repo.)

I think that all our media now has a valid source!


axelboc (approved):
Awesome!! Congrats 🙌

Issue #132, opened 2019-08-19
Flag similarities for extended deck <Country - Flag> template


axelboc:
In issue #50, there was some [confusion](https://github.com/axelboc/anki-ultimate-geography/issues/50#issuecomment-522241023) around the purpose of the current **_Flag similarity_ field**.

The field is used on the standard deck's **_Flag - Country_ template**, on the back side of the cards, to help users to **stop mixing up flags that look similar** over and over again. Typically, when a user is presented with the flag of Poland and guesses wrongly that it's the flag of Indonesia, they have the opportunity (with the help of the flag similarity information) to come up with a mnemonic to help them differentiate the two flags the next time they see either one of them.

In the case of the extended deck's **_Country - Flag_ template**, the situation is quite different. Remembering and **describing what a flag looks like** is a very different skill. From #50, it's clear that the _Flag similarity_ field is not providing the level of assistance needed for this template.

I've suggested adding a **hint field** on the front side of the cards, which could be revealed on demand by the user, to **list the countries with similar flags**. My thinking is that the hint could help jog the user's memory as to what the geometry and colours of the flags might be (as they might know what one of the other flags look like). And if they don't know what the geometry and colours of any of the flags in the group are at all, they will for sure once they see the answer and make the appropriate associations in their mind.

Am I onto something here? Is it worth adding a new hint field for the _Country - Flag_ template? Do you have any other suggestions or insights to help memorisation in the context of the _Country - Flag_ template?

aplaice:
As you could probably guess from the fact that I "misused" the similarity descriptions for this purpose, I'm broadly in favour. :)

Some possible counter-arguments:

1. From an editing perspective, adding two more fields will make dealing `data.csv` as plain text (as opposed to in a spreadsheet or pseudo-spreadsheet (like VSC's extension or Emacs's Org Mode's tables)) even more unwieldy. However, especially if #129 is fixed, this might not be a major problem.

2. Hint fields are often "abusable". (Though in this, specific case, I don't think that seeing a list of countries with similar flags beforehand would be particularly harmful to memory formation.)

3. (As you mentioned in the other thread:) There are many different potential mnemonics for cards, and I'm not sure if promoting this particular set would be generally valuable.

axelboc:
[...]

PR #469, opened 2021-09-11
Add flag similarity between Liechtenstein and Haiti

x51mon:
The flags only differ in the symbol/coat of arms and its position.

Languages other than German are auto translated.


axelboc:
@aplaice per https://github.com/anki-geo/ultimate-geography/blob/master/CONTRIBUTING.md#flag-similarity-field, I'm not convinced this pair of flags qualifies for having flag similarity information. Haiti has a _coat of arms_ of sorts in the middle, and Liechtenstein has a _symbol_ in the top left corner. Therefore, they are more different than our guidelines allow.

That being said, the ΔE of the two blues is 26.7, which qualifies as a negligible difference. So if we want to add this flag similarity pair, we could add a new critical difference to our guidelines as follows:

- decorations of different types in different positions (e.g. coat of arms in middle, symbol in top left corner)

aplaice:
@axelboc You're right!

My "reactions" were intended to indicate that this is an interesting case that's worth considering, but that I hadn't yet really thought about it or looked at the commit. (It's probably far from clear...)

FWIW I've never confused these two particular flags (and I confuse many of our flags! :)), so from my perspective our criteria do correctly identify "confusingness".

@x51mon if you're comfortable answering (and please feel free to say no), is this pair of flags a set that you've mixed up/you know somebody who mixes them up, or is it just a pair that you think "should" be on the flag similarity set?

(The similarity between Liechtenstein and Haiti _is_ IMO a good mnemonic for the Country→Flag cards, but (cf. #132) that isn't the purpose of the "Flag similarity" field.)

Hence, overall, I'd lean against changing the criteria.

x51mon:
From what you write I think I might have assumed a different purpose of this field actually. I did not yet go through all the flags but, except for a few cases, I've seen this field more as a mnemonic and/or fun fact field.

So @aplaice to answer your question: no I did not mix them up, I've just noticed that they use the same colours and found it interesting and somehow worth memorizing.

To illustrate this with another example, when going through the cards it seemed to me that Russia, Slovenia and Slovakia use (almost) the same colours but I wasn't completely sure about the order because they don't get shown next to each other. So to me that would be interesting to point out and also put on the card for the simple reason of not having to look it up. But I get now that that's not the point of this field :)

axelboc:
Right, makes sense, it is quite a confusing field! 😄 I'll close this PR then.

PR #470, opened 2021-09-12
Fix typos in country names in flag similarities

axelboc:
We've fixed typos in country names in the past but forgot to propagate the fixes to the _Flag similarities_ field.

Also fix country name for Curaçao in Russian.


src/data/flag_similarity.csv
```diff
@@ -31,11 +31,11 @@ Bolivia,Ghana (star instead of coat of arms),Ghana (Stern anstatt Wappen),Ghana
...
 Curaçao,Nauru (single star below yellow band),Nauru (einzelner Stern unter gelbem Band),Nauru (una sola estrella debajo de la banda amarilla),Nauru (une seule étoile sous la bande jaune),Nauru (en enkelt stjerne under et gult bånd),Nauru (jen jedna hvězda pod žlutým pruhem),Науру (одиночная звезда под желтой полосой),Nauru (enkele ster onder de gele band),Nauru (ensam stjärna under den gula randen)
-Nauru,Curaçao (two stars in top-left corner),Curacao (zwei Sterne in der oberen linken Ecke),Curazao (dos estrellas en la esquina de arriba a la izquierda),Curaçao (deux étoiles en haut à gauche),Curaçao (to stjerner i øverste venstre hjørne),Curaçao (dvě hvězdy v levém horním rohu),Кюрасао (две звезды в левом верхнем углу),Curaçao (twee sterren bovenaan in de linkse hoek),Curaçao (två stjärnor i övre vänstra hörnet)
```
aplaice, line 41: I think it's the Russian country name of Curaçao that should be changed to Кюрасао instead? (See: https://ru.wikipedia.org/wiki/%D0%9A%D1%8E%D1%80%D0%B0%D1%81%D0%B0%D0%BE)

axelboc: You're totally right, I wrongly assumed that @zserghei had fixed the country name in one of their previous PRs, but in fact, the inconsistency dates back all the way from the addition of the Russian translation. Thanks for double checking!!

aplaice (approved):

PR #429, opened 2021-05-13
Add more flag similarities

gaetjen:
Countries in `flag_similarity.csv` were partially ordered alphabetically, partially grouped by similarity. I suggest grouping everything by similarity to make editing and finding related entries easier. Sorting everything alphabetically could also make sense.

Added the following pairs for flag similarity (descriptions only in English and German so far. For other languages I don't know enough to add them so I only added the countries).

* Jordan, Sahrawi Arab Democratic Republic
  * Major difference: decorations of same type in different positions (star on left vs. moon and star in center)
  * Negligible difference: different geometry (Jordan has longer arrow)
* Jordan, Palestine
  * Critical difference: Presence of star (Jordan)
  * Negligible difference: different geometry (Jordan has longer arrow)
* Sahrawi Arab Democratic Republic, Palestine
  * Critical difference: Presence of moon and star (Sahrawi Arab Democratic Republic)
* Slovenia, Slovakia
  * Minor/Negligible difference: slightly different colours (Slovenia has a lighter shade of blue, is not always consistent, darker versions also exist)
  * Minor difference: different decoration of same type in same position (different coat of arms, Slovenian coat of arms slightly smaller and higher)
 * Russia, Slovakia
   * Critical difference: presence of coat of arms
 * Russia, Slovenia
   * Critical difference: presence of coat of arms

Did NOT include:
* Sudan with Sahrawi Arab Democratic Republic or Jordan
  * 1 critical difference (presense absence of decoration) + 1 major difference (rotated colors)

PR #473, opened 2021-09-18
Add flag similarities for Russia, Slovakia, Slovenia and Palestine, SADR

axelboc:
For some reason, GitHub closed #429 when I merged #470 instead of changing the base branch back to `master`... So, here it is again, starting from my last comment!

---

The case of Jordan/SADR is interesting. The major difference "decorations of same type in different position" was initially meant to care only about the type of decoration (i.e. symbol vs. coat of arms vs. glyph vs. ...), and not the decorations themselves. So two symbols appearing in different positions was supposed to fit the rule, regardless of the symbols themselves.

That being said, from a subjective point of view, I agree with you @aplaice that the symbols of Jordan and SADR are different enough to not qualify for major difference status. If the stars had the same number of branches, maybe ... Anyway, so perhaps we should clarify the guidelines somehow?


src/data/flag_similarity.csv
```diff
@@ -39,3 +40,6 @@ Guinea,"Mali (red and green flipped, slightly brighter green)","Mali (Rot und Gr
...
+Russia,"Slovakia (with coat of arms),"Slowakei (mit Wappen),"Eslovaquia (con escudo)","Slovaquie (avec blason)","Slovakia (med våpenskjold)","Slovensko (s erbem)","Словакия (с гербом)","Slowakije (met wapen)","Slovakien (med vapensköld)"
+Slovakia,"Russia (no coat of arms), Slovenia (wider, smaller coat of arms)","Russland (kein Wappen), Slowenien (breiter, kleineres Wappen)","Rusia (sin escudo), Eslovenia (más ancha, escudo más pequeño)","Russie (pas de blason), Slovénie (plus large, blason plus petit)","Russland (uten våpenskjold), Slovenia (bredere, mindre våpenskjold)","Rusko (bez erbu), Slovinsko (širší, menší erb)","Россия (без герба), Словения (шире, меньший герб)","Rusland (geen wapen), Slovenië (breder, kleiner wapen)","Ryssland (ingen vapensköld), Slovenien (bredare, mindre vapensköld)"
+Slovenia,"Slovakia (narrower, bigger coat of arms)","Slowakei (schmaler, größeres Wappen)","Eslovaquia (más estrecha, escudo más grande)","Slovaquie (plus étroit, blason plus grand)","Slovakia (smalere, større våpenskjold)","Slovensko (užší, větší erb)","Словакия (уже, больший герб)","Slowakije (smaller, groter wapen)","Slovakien (smalare, större vapensköld)"
```
axelboc, line 45: - Technically narrower should go last because it's a minor difference, but it felt clearer in first position.
- I chose not to mention the position of the coat of arms, as I felt like its size difference was significant-enough on its own and more concise to describe.

src/data/flag_similarity.csv
```diff
@@ -22,7 +22,8 @@ Sudan,"Palestine (black/white/green, red arrow)","Palästina (Schwarz/Weiß/Grü
...
+Sahrawi Arab Democratic Republic,"Palestine (no symbol)","Palästina (kein Symbol)","Palestina (sin símbolo)","Palestine (pas de symbole)","Palestina (uten symbol)","Stát Palestina (bez symbol)","Палестина (без символа)","Palestina (geen symbool)","Palestina (ingen symbol)"
```
axelboc, line 26: I don't think we need to mention "star and crescent" here -- the term "symbol" seems clear-enough.

aplaice (approved):
Thanks very much for finishing this off!

I've read through the changes, and everything seems correct. (I haven't cross-referenced all the terms in the various languages with their previous occurrences, but I assume that if the translations are slightly wrong, we'll eventually get feedback!)

PR #474, opened 2021-09-23
Improved Russian translation

zserghei:


aplaice:
Should we merge this? (Is it ready?)

Merging will need to be done very carefully to ensure that we don't overwrite either set of changes. Would the following set of steps (assuming the PR is otherwise ready) be sensible:

1. I'll manually merge master into this branch.
2. I'll push this to this branch (I think I have the needed permissions).
3. We can triple-check that everything is OK before merging into `master` via GitHub.

?

(If we/GitHub had a suitable mergetool for CSV files this merge would be trivial. :/)

axelboc:
daff looks amazing :star_struck:

aplaice:
> daff looks amazing 

It seems to be amazing — the merge of `master` into this branch was done 99% automatically. (The only manual change was removing NULLs from the Portuguese fields in Namibia and the English Channel in `country_info.csv`, which was necessary because `daff` wanted to emphasise that the fields were absent from both sources (they were in a new row in one and in a new column in the other).)

(I double-checked using `csvtool` that only the Russian columns were changed.)

> Would the following set of steps (assuming the PR is otherwise ready) be sensible:

I've done 1. and 2.  I'm not sure when we should merge (soon? or wait for more fixes?), but I think everything is OK from a technical PoV.

<hr/>

@zserghei Thanks again very much for all these fixes and for the thought you've put into it (e.g. adding country info where it makes sense for Russian, even when it didn't make sense in other languages)! :) Would you mind if we added you as [(one of?) the translator(s) for Russian](https://github.com/anki-geo/ultimate-geography/wiki/Translators)?

aplaice (approved):

zserghei:
@aplaice Yes, you may add me as the translator for Russian.

aplaice:
> @aplaice Yes, you may add me as the translator for Russian.

Great! Thanks!

<hr/>

I assume that we can merge(?). (AFAICS everything is ready.)

zserghei:
I think that it can be merged now.

PR #476, opened 2021-09-26
Replace "Downloads" with "Assets" in installation instructions

horwitz:
It seems like the installation instructions are talking about the section labeled "Assets" (when it says "Downloads"); this PR is to update that wording.


axelboc (approved):
Good catch 🎣

PR #479, opened 2021-09-28
Clarify major version upgrade instructions in README

axelboc:
As discussed here: https://github.com/anki-geo/ultimate-geography/discussions/475#discussioncomment-1387060, I've reworded the instructions to make it clear that we typically provide a way to upgrade without losing progress.

I've also updated all the old repo URLs in the docs and deck description.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 :wave: Hello there! :tada: Thanks for taking the time to contribute!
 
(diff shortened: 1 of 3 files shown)


aplaice (approved):
Looks great! I hope this will minimise the number of people who upgrade by resetting their progress when they don't want that to happen.

(I think that the previous version was already relatively clear, but this is even better!)

PR #482, opened 2021-10-02
Make map of Ukraine more consistent

### LINKED ISSUE #471: Map of Ukraine is the only one showing meridians and circles of latitude
This makes it easier to use that as a trick to associate the image with Ukraine without actually paying attention to the country.

I plan to fix this myself as it's very easy, but I don't have very much time now, so feel free if you'd like to solve it. ^^

josealberto4444:
This commit removes the meridians and circles of latitude from the map
of Ukraine.

I don't think it's a change significant enough to add it to Wikimedia, so I just added the difference in sources.csv.

Fixes #471.


axelboc (commented):
Cool, thanks! Could you please just keep the dimensions as they were (i.e. 500px width like the other maps)?

josealberto4444:
Oops. Sorry, I forgot to check that. 😅 I just used 96 DPI and obtained a number that sounded familiar to me. =P

I'll update it today. =)

aplaice (approved):
Thanks so much!

PR #489, opened 2021-10-24
PT Complete Translation

KitsuneModder:


KitsuneModder:
I can do whatever you find better ;)

Just decide and tell me :)

aplaice:
It's hard for either @axelboc or me to decide, since (AFAIK) neither of us speaks Portuguese, but based on the discussion I think that `Copenhague (BR), Copenhaga (PT)` (in the `Country` and `Capital` fields) is the most accepted option. (Ideally with an explanation in a user-facing document (deck description? the Readme?) and optionally some JS to prettify things, but we can think about both of those later — IMO it's just about understandable and presentable without them.)

We didn't explicitly discuss what to do with mentions of the relevant countries in the Flag similarities (and the one mention of Macedonia in the country info (where the old name is the same as the new name, just without "North")), but I think that there having `Letônia/Letónia` (as in the current version of the PR) is least confusing/annoying, on the whole.

axelboc:
I came across this article, which explains a fair bit about the history of Portuguese Wikipedia and the way it works with regard to dialects, notably this:

> It is widely accepted that an article written in Brazilian Portuguese should be kept as it is, and the same applies to articles in European Portuguese.

This, I think, confirms my initial assumption that the differences are well-known to all native speakers and don't impede readability.

Having both BR and PT in the _Country_ and _Capital_ fields is still my recommendation -- even with BR first like you suggested @aplaice - i.e. `Copenhague (BR), Copenhaga (PT)` -- but I reckon we can safely decide to use the Brazilian dialect only for every other field, including for country names in the _Flag similarity_ field.

---

So as far I see it, here's what's left to do before merging:

- [ ] Remove the remaining articles from the _Flag similarity_ field.
- [ ] Change the format of the country and capital names with dialect variations from `Copenhague/Copenhaga` to `Copenhague (BR), Copenhaga (PT)`

We can do the following after the PR is merged:

- [ ] Document the new translation in the README and deck description
- [ ] Add a _Portuguese_ section to the [_Translation sources_](https://github.com/anki-geo/ultimate-geography/blob/master/CONTRIBUTING.md#translation-sources) section of the contributor's guide to document the decisions above, if you both agree with them.
- [ ] Consider adding some JS to the templates to reduce the visual impact of parenthesized content in the _Country_ and _Capital_ field. This could actually be beneficial to the other rare instances of parens in these fields, notably the disambiguation texts for Saint Martin and Sint Maarten.

src/data/capital.csv
```diff
@@ -34,19 +34,19 @@ Luxembourg,Luxembourg City,Luxemburg,Luxemburgo,Luxembourg,Luxembourg,Lucemburk,
 North Macedonia,Skopje,Skopje,Skopie,Skopje,Skopje,Skopje,Скопье,Skopje,Skopje,Escópia
 Malta,Valletta,Valletta,La Valeta,La Valette,Valletta,Valletta,Валлетта,Valletta,Valletta,Valeta
 Moldova,Chișinău,Chișinău,Chisináu,Chișinău,Chisinau,Kišiněv,Кишинэу,Chișinău,Chișinău,Quixinau
-Monaco,Monaco,Monaco,Mónaco,Monaco,Monaco,Monaco-Ville,Монако,Monaco,Monaco,Mônaco/Mónaco
+Monaco,Monaco,Monaco,Mónaco,Monaco,Monaco,Monaco-Ville,Монако,Monaco,Monaco,"Mônaco (BR), Mônaco (PT)"
```
aplaice, line 37: Should this be?

```suggestion
Monaco,Monaco,Monaco,Mónaco,Monaco,Monaco,Monaco-Ville,Монако,Monaco,Monaco,"Mônaco (BR), Mónaco (PT)"
```

aplaice:
Other than the one tiny typo, this looks great!

I'd vote for waiting for @axelboc to have a final look and then merging asap! :)

axelboc (approved):
Looks great to me, thank you so much!!

PR #491, opened 2021-11-05
Fix CSV delimiter typos

aplaice:
Oops! Sorry for not catching these earlier!

(Aside: apart from issues caused by these, CSV-merging the Russian fixes branch into master with daff seems to work great! :))


axelboc (approved):
🎣

PR #506, opened 2022-01-17
Update flag of Eswatini

josealberto4444:
It appears to have changed in 2011 and Wikipedia updated it. There's no
need to change sources.svg as the file it references is the one I'm
committing here.

I used [svgo](https://github.com/svg/svgo), then [scour](https://github.com/scour-project/scour/) and then i trailed the newlines using just `tr`, but the file is still bigger than the previous one (6.3 KiB vs 6.6 KiB). Did you compress svg files differently? I can amend the commit with no problem. =)


aplaice:
Thanks for noticing! :)

<details><summary>Resolved doubts about flag update</summary>

I'm slightly hesitant to update given that the official government [Eswatini websites](https://web.archive.org/web/20220101043818/http://www.gov.sz/) (unarchived version currently seems down) and [web accounts](https://twitter.com/EswatiniGovern1) still use the old version of the flag and the "original source" used by FOTW was a South African site (that's now down and remains unarchived), as pointed out in the [talk page](https://commons.wikimedia.org/wiki/File_talk:Flag_of_Eswatini.svg).  (We don't want to make changes if it's likely that Wikipedia/Wikimedia will revert again.)

However, Britannica does have the [updated flag](https://www.britannica.com/topic/flag-of-Swaziland) and the image sources ("Sources for Tassels" on https://commons.wikimedia.org/wiki/File:Flag_of_Eswatini.svg) do indicate that in physical use, the government has indeed updated the flags.  Also, googling for images for "kingdom of eswatini press release" yields some more flags with gray or black tassels, and none with the old ones.

Hence, I think we're safe to change!

</details>

<hr/>

The file size increase is likely because the source image increased in size (from 6 kB to 9 kB) and svgo isn't quite as good as whatever FDRMRZUSA did on Wikimedia.  (Yes, we use svgo, but relatively frequently the images on Wikipedia were optimised already.)  I don't think that the 300 byte increase is worth worrying about.  If you're determined to cut it down, perhaps remove the XML "preamble" (I think it's not essential) and remove spaces between tags (`> <` etc.)?

<hr/>

Please change the SVG height to 250 (and adjust the width proportionally), for consistency, though!

josealberto4444:
Yes, I also saw the edit war in Wikimedia, but I thought that it's just a matter of time that it settles definitely with the grey tassels, too. =)

> Please change the SVG height to 250 and adjust the width proportionally, (just in the top element), for consistency, though!

I think it's done. =)

josealberto4444:
It wasn't, but it should be done now, hopefully. =P

axelboc (approved):
Looks great, thanks!

PR #507, opened 2022-01-18
Add zoombox to Gibraltar map

### LINKED ISSUE #461: Add Gibraltar zoombox (suggestion)
Do you think it's a good improvement for the deck? I know there is no ambiguity in this case, but right now, in my opinion, the map is too vague.

I've been taking a look to #357 to try making the zoombox for Gibraltar and see if it looks well, but I don't really know what's the procedure you followed for making these zoomboxes. Could you explain it to me, @aplaice? =)

josealberto4444:
Closes #461.


axelboc (approved):
:heart:

aplaice:
Looks amazing! :)

Please upload it to Wikimedia (as something like "Gibraltar in its region (zoom).svg"), so that other people can also benefit from your amazing work! (Also, in that case, please change `sources.csv` to point to your version.)

josealberto4444:
Yes, I was going to write about that in the issue after making the PR and then I forgot. Oops. =P

Thanks!

PR #508, opened 2022-01-18
Change name of capital as adapted to Spanish

josealberto4444:
Finally there is consensus about how to write this capital, as the two
official language institutions (RAE and Fundeu) pronounced publicly
about it.

I know that this conflicts with the Chinese translation and will have to wait until it's merged, but I wanted to leave it here so I don't forget. ^^


aplaice (approved):
Thanks!

(As you wrote, we'll wait for the Chinese translation to be merged before merging this, to avoid making that more complicated.)

Issue #519, opened 2022-02-14
A Yellow Sea bug and a request to add Bohai Sea


ruin1990:
I found that the map of the Yellow Sea in Deck was wrong, it included the Bohai Sea. In fact, the Bohai Sea and the Yellow Sea are two different seas, and the color and composition of the sea water are also different.

I want to help correct this problem, but I don't know how to draw the map in Ultimate Geography. 
@aplaice @ohare93,  can you help it?

Map for Yellow Sea From Ultimate Geography
![ug_yellow_sea](https://github.com/anki-geo/ultimate-geography/blob/master/src/media/maps/ug-map-yellow_sea.png?raw=true)

Map for Bohai Sea from wiki：
![bohai_sea](https://upload.wikimedia.org/wikipedia/commons/6/67/Locatie_Bohaizee.PNG)
Map for Yellow Sea from wiki：
![yellow_sea](https://upload.wikimedia.org/wikipedia/commons/8/8f/Locatie_Gele_Zee.PNG)

Reference: 
https://en.wikipedia.org/wiki/Bohai_Sea
https://en.wikipedia.org/wiki/Yellow_Sea
https://zh.wikipedia.org/wiki/%E9%BB%84%E6%B5%B7
https://zh.wikipedia.org/wiki/%E6%B8%A4%E6%B5%B7

aplaice:
Thanks for the bug report.

However, according to all the English-language sources I could find, the Bohai Sea is part of the Yellow Sea.

> The innermost bay of northwestern Yellow Sea is called the [Bohai Sea](https://en.wikipedia.org/wiki/Bohai_Sea) ([Wikipedia article on the Yellow Sea](https://en.wikipedia.org/wiki/Yellow_Sea))

> It is the northwestern and innermost extension of the [Yellow Sea](https://en.wikipedia.org/wiki/Yellow_Sea), ([Wikipedia article on Bohai Sea](https://en.wikipedia.org/wiki/Bohai_Sea)

> The Yellow Sea [...] includes the Bo Hai and Liadong Hai ([IHO 2002 draft](http://wiki.geosys.ru/lib/exe/fetch.php/ru/portal/lib/iho/s23.los.ed4draft.2002.pdf#page=164)

[IHO 1953](https://epic.awi.de/id/eprint/29772/1/IHO1953a.pdf#page=33) doesn't mention Bohai Sea.

[...]

PR #524, opened 2022-02-22
The Chinese version adds an explanation of the Bohai Sea to the description of the Yellow Sea (#519)

ruin1990:
The Chinese version adds an explanation of the Bohai Sea to the description of the Yellow Sea (#519)

ps: I have a question, why every PR, this commit information is getting more and more


aplaice (approved):
Thanks!

<hr/>

> ps: I have a question, why every PR, this commit information is getting more and more

The issue is that your first PR was "squashed" — all your changes were combined into a single commit (for neatness).  This means, that while all the content of all your changes was included, into `anki-geo`, `git` doesn't fully know that (approximately speaking, it tracks history via commit "chains"/graphs, so since the commit graphs are different, it thinks that something is missing).

If you reset your branch to anki-geo `master`, then the issue would be resolved (this is what I had previously suggested, but I wasn't too clear — sorry!

For simplicity, please wait for this PR to be merged. For instance, if you're using command-line git:

```bash
git branch old_master master # Not really needed — create a "backup" of your old branch, just in case. (You can also access all old versions of the branch with `git reflog`, but that's more complicated).
git fetch https://github.com/anki-geo/ultimate-geography/ # Fetch the anki-geo master branch
git reset --hard FETCH_HEAD # Reset your master to anki-geo master
git diff old_master master # Make sure that the actual contents is the same, as before.
# git branch -D old_master # If you don't need the old "version" of the master branch (which you probably don't).
```

If this doesn't make sense, please ask or [see the git book](https://git-scm.com/book/en/v2).

PR #527, opened 2022-03-07
Add alternative Dutch name to Belarus

flipflop97:
In the Netherlands, Belarus is considered an alternative (and in context of the current war almost primary) name to "Wit-Rusland". The [Dutch Wikipedia page](https://nl.wikipedia.org/wiki/Wit-Rusland) also states both names, so it would make sense to add this alternative and internationally used name.


aplaice (approved):
Yes, definitely! Thanks!

PR #529, opened 2022-03-07
Change "gekend" to "bekend" in Dutch country info

flipflop97:
The word "gekend" is incorrectly used here, and should be "bekend"


aplaice (approved):
Thanks! (Again, I can only verify via Wiktionary, googling of phrases and google translates, but to the extent I can tell this is probably the right change. (I tried to at least vaguely check, otherwise it'd mostly be a case of whose Dutch we trust more: yours or the original translator's. :)))

PR #530, opened 2022-03-07
Change "Beweerd" to "Geclaimd" in Dutch capital hint

flipflop97:
The word "Beweerd" is a translation of "Claimed" in another context, in this context "Geclaimd" is more accurate. It is also more consistent, as it is used in the translation for the hint of Israel.


aplaice (approved):
Thanks for these fixes!

PR #536, opened 2022-04-21
Remove 'European_Union' tag from United Kingdom

moritzhertler:
The UK has not been part of the EU since 2020. See [Brexit](https://en.wikipedia.org/wiki/Brexit).


aplaice (approved):
Oops! We had updated the EU map immediately (after the UK actually left), but not removed the UK's tag...

Thanks!

PR #537, opened 2022-04-23
Update to BrainBrew 0.3.8

### LINKED ISSUE #502: Upgrade to latest BrainBrew
This should hopefully be straightforward, if not completely trivial. (See #498 for potential errors.)

It might be useful, so that potential fixes (like for encoding issues) can be easily propagated to us as well.

<hr/>

I'll have a look soon-ish. I'm opening an issue to keep track and in case anybody has any comments.

aplaice:
Close #502.

I think that conversion both ways seems to work, now.

The recipes might not be the most idiomatic BrainBrew recipes, but I think they're correct.

AFAICT the contents of the output with `source_to_anki` with BrainBrew 0.3.6 is the same as with BrainBrew 0.3.2 (what we had been using until now), except that the deck JSONs are now all called `deck.json` (in line with the default changed name upstream in CrowdAnki) and that a couple of weird fields (`bfont`, `bsize`) have now been added (but AFAICT they do potentially have a useful function).

Import of the generated CrowdAnki decks seems to work correctly!

<hr/>

This should probably wait for the Polish translation to be merged, first!

<hr/>

I haven't implemented the switch to TSVs — we can do that later, if we want to!

<hr/>

For context, upgrading to latest BrainBrew is valuable, because it allows us to avail ourselves from all  improvements and bug fixes (e.g. hopefully the content encoding issues — if we've actually solved them...) that have been added to BrainBrew.


aplaice:
I'm squashing and merging, since AFAICT everything works.

If @ohare93 you have any improvement suggestions (which are now possible due to the newer BrainBrew!), we can incorporate them later on!  (I hope I didn't make too much of a mess. :/)

PR #547, opened 2022-06-28
Updated German Translations for two Country Names

pdeubel:
Updated the german translations for 

- Canary Islands
    - Was "Kanarischen Inseln", but should be "Kanarische Inseln" [Wikipedia DE](https://de.wikipedia.org/wiki/Kanarische_Inseln)
- Sahrawi Arab Democratic Republic
    - Was "Demokratische Arabische Republik Sahara (DARS)", but should be "Demokratische Arabische Republik Sahara" [Wikipedia DE](https://de.wikipedia.org/wiki/Demokratische_Arabische_Republik_Sahara)


aplaice (approved):
Thanks!

(Oops, you're right — it's "die Kanarischen Inseln" (with the article) but "Kanarische Inseln" (without the article).)

PR #555, opened 2022-07-23
[PL] Fix capital of Federated States of Micronesia

giovanni-cutri:
The name of the country was incorrectly reported as the capital.


axelboc (approved):
Good catch 🎣

PR #557, opened 2022-07-26
[FR] Fix wrong info about Kazakhstan capital

giovanni-cutri:
Just a small detail, the name of the month when the capital changed its name was wrong.


axelboc (approved):
🎣

PR #566, opened 2022-08-22
Treat Spanish and Chinese decks as distinct "objects"

### LINKED ISSUE #565: Importing Spanish (or Chinese) deck over English deck changes the deck description
As described [here](https://github.com/anki-geo/ultimate-geography/discussions/564#discussioncomment-3448043), if you import the Spanish deck over the English deck you end up with:

1. One single "Ultimate Geography" deck in Anki.
2. whose deck's description is in Spanish.
3. which contains all the cards/notes twice, once in English once in Spanish.

1 and 3 were intentional when we decided to use the same `crowdanki_uuid` for all decks. 2 was not. (It happened because in the meantime we had added a localised Spanish deck description.)

(The same holds for the Chinese deck, but no other decks.)

This is confusing, because the deck description will be different depending on the order of import (the language of the deck description will be that of the last deck imported).

It also breaks the idea that the deck UUID uniquely identifies a distinct deck "object".

<hr/>

We can easily fix it by changing the `crowdanki_uuid` for the Spanish and Chinese decks. (We'd have to have a separate `src/header/default_es.yaml` and `src/headers/default_zh.yaml`.) (Overrides don't work with the `crowdanki_uuid` and it'd be tricky to make BrainBrew work with them correctly (we'd probably have to have a `crowdanki_uuid_yaml_file` in the overrides so that both reading and writing of the UUID would work).)

aplaice:
This arguably makes sense, since they have translated deck descriptions (unlike the other language decks).

Fix #565.

I'm not too happy about this because:

1. We're repeating ourselves considerably (most of `src/headers/default.yaml` is repeated).
2. While repeating ourselves we're duplicating the definition of the "Ultimate Geography" deck config (`43d37fac-9a65-11e8-9cc5-a0481cc15658`), which IMO should still be exactly the same for all decks.
3. We now have "separate" decks for only Chinese and Spanish (because these are the only decks with separate deck descriptions).  (This arguably makes sense, and I'm least unhappy about it.)


ohare93:
@aplaice this change should help lighten the duplication :+1: https://github.com/ohare93/brain-brew/pull/47

With this change you do not need an extra header yaml for the different languages, but can instead read in the same yaml file multiple times, with overrides for the description (which was already possible), name, crowdanki_uuid.

It took me 10 mins to do, but I did not get time to test it. It's pretty simple though :smile: Let me know if it works or not, and I can take a look tomorrow at the latest.

Pipfile
```diff
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-brain-brew = "==0.3.8"
+brain-brew = {git = "https://github.com/ohare93/brain-brew.git", editable = true, ref = "da3089a0c0633fe86bc6a3430979944babf23a47"}
```
ohare93, line 9: I will hit the release button whenever you like, so this can be a proper release version :grin:

aplaice: Yeah, this wasn't intended to be for merging yet (just in case somebody else wanted to easily test)! :)

recipes/source_to_anki.yaml
```diff
@@ -28,14 +28,20 @@
...
         override:
           deck_description_html_file: src/headers/desc.html
+          crowdanki_uuid: 43c5ba66-9a65-11e8-90c9-a0481cc15658
+          name: Ultimate Geography
       - part_id: default header:es
         file: src/headers/default.yaml
         override:
           deck_description_html_file: src/headers/desc_es.html
+          crowdanki_uuid: cb4d32ee-12ed-9960-1841-28c09449ded0
+          name: Ultimate Geography [ES]
```
axelboc, line 38: Should we translate the names as well, then? 🤷

axelboc:
Congrats on fixing (1) and (2)!

Re. (3), I think having separate decks makes sense, and seems to be what people expect (cf. https://github.com/anki-geo/ultimate-geography/discussions/564#discussioncomment-3448043). Sure, it's annoying that not all translated decks are separate at the moment, but it's temporary. If we agree that this is the way to go for future translations, then I think it's good to go!

recipes/source_to_anki.yaml
aplaice: As in, say, "Geografía definitiva"? :D 

Interesting!

I'd lean towards no, at least for now:

1. It might make explaining how to deal with the deck _slightly_ longer/_slightly_ more confusing.

    Now, we can just say that the deck is "Ultimate Geography", and even if it'll be "Ultimate Geography [ES]", it'll still be clear what we're referring to. If it'll be translated in some languages, but not others, we might need a long-winded explanation or a list, to avoid confusion.

2. For those using multiple language decks having several "Ultimate Geography ([x])" feels neater.
3. Note types are still not localised (we still have "Capital", "Country" etc.) and IMO deck names should be the last thing to be translated.

axelboc (approved):
🎉

PR #570, opened 2022-09-23
Expand SADR country info mentioning alternative names

### LINKED ISSUE #561: SADR vs. Western Sahara (in the Polish translation and elsewhere)
Sorry for spinning this out into a separate issue out of #559, but I'm not quite sure what's the "right thing to do" here, and the scope is slightly larger than I initially thought.

Our Polish translation of SADR is currently "Sahara Zachodnia" (i.e. Western Sahara).  This is arguably correct, since the only Polish Wikipedia article about the "area" (nebulously defined) has "Sahara Zachodnia" as the title.

English Wikipedia (and most other Wikipedias, but apparently not Polish Wikipedia) have two separate articles for "the geographical area" of Western Sahara (currently mostly under Moroccan control/occupation, but some under that of SADR) and for the partially recognised state of SADR (whose claimed borders coincide with those of Western Sahara).

(The English Wikipedia article about [SADR](https://en.wikipedia.org/wiki/Sahrawi_Arab_Democratic_Republic) corresponds/links to https://pl.wikipedia.org/wiki/Saharyjska_Arabska_Republika_Demokratyczna which redirects to the article [Sahara Zachodnia](https://pl.wikipedia.org/wiki/Sahara_Zachodnia).  The English article about [Western Sahara](https://en.wikipedia.org/wiki/Western_Sahara) directly corresponds/links to [Sahara Zachodnia](https://pl.wikipedia.org/wiki/Sahara_Zachodnia).)

Hence, we could argue that there are two separate concepts (the "geographical area" and the partially recognised state) and for the Polish deck we're translating a different one than in all the other languages (and hence that the main country name in the Polish deck should be "Saharyjska Arabska Republika Demokratyczna").

However, even English Wikipedia says that SADR is "also known as Western Sahara" (i.e. "Western Sahara" can both be the name of the area and the state).  Also, to look at a vaguely similar case, for the UK we use the "Great Britain" name (i.e. the translation of GB) for the UK in the Norwegian, Swedish, Russian and Polish decks (Storbritannia etc.), despite the "United Kingdom" name also existing in these languages (Det forente kongeriket etc.) (and despite "Great Britain" also being the name of the island), because the articles about the country in the relevant languages use the "Great Britain" name as the article title (which we assume is because the "Great Britain" name is the one most frequently used in these languages).  This would probably suggest that we should keep the Polish name as "Sahara Zachodnia".

<hr/>

Overall, I think that:

1. We should keep all the SADR names in `country.csv` as they are.
2. In the English, Czech, Swedish and Italian decks, mention "Western Sahara" in the country info (as an alternative name?).  (These are the languages whose Wikipedias AFAICT say that SADR is also known as "Western Sahara" (or similar phrasings).)  (Swedish already does, so it wouldn't need changing.)
3. (?) In the Polish deck, mention "Saharyjska Arabska Republika Demokratyczna" in the country info (not having SADR mentioned would IMO be weird).  However, how should it be mentioned?

   As an alternative name?  (Also known as...)

   Polish Wikipedia doesn't actually say the equivalent of "Western Sahara is also known as SADR".  Also, saying that "Western Sahara is also known as SADR" is ambiguous, since Western Sahara (the area) is not also known as SADR (AFAICT in any language), even if Western Sahara (the state) is also known as SADR.
4. (?) In the remaining languages don't mention "Western Sahara".

aplaice:
Fix #561.

As discussed in #561, saying that "Sahara Zachodnia" (Western Sahara) is also known as SADR (in the Polish version), is ambiguous and potentially misleading, since Western Sahara is both the name of the geographic area (ignoring political associations) and one of the names used for the partially recognised state.  However, AFAICT, we have the exact same situation for Artsakh, in Polish, (Górski Karabach (Nagorno-Karabakh) also known as Artsakh, where Nagorno-Karabakh can refer both to the country and the geographical area.).  Also, since we're generally dealing with countries, here, it should be clear that we mean Western Sahara (the country) rather than WS (the area), so I think my initial worry was overblown.

I'm not 100% sure if I have the correct gender for conosciuto in Italian (should it agree with Repubblica.. or with stato?).

The same question holds for Czech — should známý agree with ..republika or with stát?

I've gone with agreement with stát for Czech (male) and with republic (female) for Italian, because in the former the second clause is separated only with a comma, while in the latter it's separated by a semicolon.  The choice of separator was based on precedence in other cases (Faroe islands and Taiwan).


axelboc (approved):
Nicely done!

PR #571, opened 2022-09-23
Make Astana capital again

KentGrigo:
The capital of Kazakhstan has been changed back from Nur-Sultan to Astana as of September 2022: 
https://www.theguardian.com/world/2022/sep/14/kazakhstan-to-change-name-of-capital-from-nur-sultan-back-to-astana

As the names just needed to be swapped, I tried to change it for all the languages. 
But it'd be nice if someone with more knowledge about these various languages could double check the correctness.


aplaice (approved):

aplaice:
Oops; checking the translations (I thought that I had checked them previously and hence only looked through cursorily today), I think that the Spanish and Portuguese translations are wrong, with an unnecessary `de`. (The `de` was in the previous translation as part of `marzo de 2019` ("March of 2019") and similarly for Portuguese.)

The Czech translation also has an unnecessary/wrong month (`březen` is March).

Issue #573, opened 2022-10-04
Typo in Dutch name for "Wallis and Futuna"


reinpeter:
In Dutch it says Wallis en Fortuna, where it should be Wallis en Futuna.
See: https://nl.wikipedia.org/wiki/Wallis_en_Futuna

PR #575, opened 2022-10-05
issue-573 fix type on Wallis en Futuna

reinpeter:
Fix: #573 
NB, some trailing space got fixed too automagically (as required by .editorconfig)


aplaice (approved):
Thanks!

PR #576, opened 2022-10-05
issue-574 fix spelling error on koniNkrijk

reinpeter:
Fix: #574 including other entries with the same spelling error

NB, some trailing space got fixed too automagically (as required by .editorconfig)


aplaice (approved):
Thanks very much for catching this pervasive typo! Looks great!

PR #577, opened 2022-10-08
Vectorize flags

KentGrigo:
This PR will vectorize the following flags:
- Andorra
- Croatia
- Dominican Republic
- Ecuador
- Guatemala
- Mexico
- Montenegro
- San Marino
- Serbia

--------------------

The following flags still need to be vectorized:
- Belize
- Bolivia
- Costa Rica
- El Salvador
- Peru
- Venezuela

These flags were not vectorized because they also needed to be blurred:
- Bolivia
- El Salvador

and these flags because they didn't correspond to the first flag on Wikipedia, which I thought was the consensus based on https://github.com/anki-geo/ultimate-geography/issues/111:
- Belize
- Bolivia
- Costa Rica
- Peru
- Venezuela


axelboc:
Awesome, thanks!! Did you run them through [SVGO](https://jakearchibald.github.io/svgomg/)? Otherwise, maybe @ohare93 can run them through [ImgBot](https://github.com/marketplace/imgbot) in his fork 😁.

KentGrigo:
I have now. 😅 Thanks for reminding me.
I'm not familiar with SVGO so I didn't change any of the settings, except for Croatia's flag.
The file size would have been +18,34% with the default settings. 
I disabled `Move group attrs to elements` as it changed it to -1.5%.

axelboc (approved):
They look much better than the old PNGs, thanks 🤩

aplaice:
Thanks!

I think that these should, ideally, have [their height changed to 250 px and the width adjusted accordingly](https://github.com/anki-geo/ultimate-geography/blob/master/CONTRIBUTING.md#flag-field). (Otherwise, the images will appear huge when reviewing.)

Just editing `height` and `width` in the `svg` element should be sufficient.

(The contributing guidelines also mandate having the `viewBox` attribute, but I'm not sure if it's actually necessary — or rather I'm not sure what adverse effects its lack currently causes.)

aplaice:
Thanks!

I think that these should, ideally, have [their height changed to 250 px and the width adjusted accordingly](https://github.com/anki-geo/ultimate-geography/blob/master/CONTRIBUTING.md#flag-field). (Otherwise, the images will appear huge when reviewing. (Modifying the SVG as opposed to specifying the height with CSS has, I think, the advantage that the images will also look consistent (and not huge) in the editor/browsing pane.)

Just editing `height` and `width` in the `svg` element should be sufficient.

I think that the `viewBox` attribute also needs to be added.

<hr/>

One other very minor issue (which I realised while looking at #220) is that now, if one zooms in, the name of the Dominican Republic is visible. Hence, in the longer run, it might be also worth blurring.

KentGrigo:
@aplaice, thank you for correcting the widhts and heights, plus adding the viewBoxes.

PR #578, opened 2022-10-10
Resize flag of Indonesia and add viewBox to flag of Eswatini

aplaice:
I think that the viewBox for Indonesia is needed, while it doesn't seem to be for Eswatini (maybe because the "internal" size is the same as the "external" size?).  I've added it in to Eswatini for consistency.

In other words: the image for the flag of Indonesia definitely does need fixing, while that for Eswatini probably doesn't, but the change doesn't hurt.


axelboc (approved):
Thanks!

PR #580, opened 2022-10-23
Add viewBox, set height to 250 and adjust width proportionally

aplaice:
(For just-added flags.)

See:
https://github.com/anki-geo/ultimate-geography/pull/577#issuecomment-1272623940

(I would have preferred these to have been fixed before merging, for the very marginal benefit that then the unfixed versions of the files wouldn't be in the main git history and hence could eventually be garbage-collected, but given that these are SVGs and hence can be compressed relatively well in the pack files (and even if they couldn't it's ≲1 MB), it's probably not worth over-worrying about.)


axelboc (approved):
Thanks!

PR #582, opened 2022-10-31
Add extra info for Sea of Galilee in dutch

reinpeter:
Added alternatives in country info field for "Meer van Tiberias".
The other ones (see https://nl.wikipedia.org/wiki/Meer_van_Tiberias) are well known either internationally or from biblical stories.


aplaice (approved):
Thanks!

PR #583, opened 2022-10-31
Differentiate better between state and region of micronesia

reinpeter:
In Dutch, Micronesia state and region have same name (anki even shows 'duplicate').
Therefore, country is changed to 'Fed state of M', like most other languages. This is also the official name (as stated on https://nl.wikipedia.org/wiki/Micronesi%C3%AB_(land)), even though the page title does not reflect that. 
The short forms are now mentioned in country info field.


aplaice (approved):
Thanks for catching this!

Nitpick: we aim to end all "country info" phrases with a full stop, for consistency.

I'm not sure if we need both `Micronesië` and `Micronesia` in the country info — not that they hurt! (I vaguely recall us having decided to aim to only include alternative names if they're different enough, to avoid overload/confusion for the user, but I can't find the reference.)

axelboc:
Thanks for all the fixes @reinpeter!

Let us know if you think both alternative names are worth keeping or if one is significantly more used than the other.

reinpeter:
The one with the ië is more used.
But I also think it is ok to keep both, the ia version is more like the international (EN) version, which is always good to be exposed to too.

PR #585, opened 2022-10-31
Add alternative for capital of China

reinpeter:
Add alternative for China capital.
The name "Peking" (or similar) is used in some languages as official name for Beijing.
I think it is important enough to know the official Chinese name 'Beijing' too, since it is also used internationally.


aplaice (approved):
Makes sense, thanks!

(Again, nitpick regarding final full stop.)

PR #586, opened 2022-10-31
Fix capitalization of Eswatini

reinpeter:
Although some spellings seem to use capitalization in the middle of the word (?), the Dutch name does not.
https://nl.wikipedia.org/wiki/Swaziland


aplaice (approved):
Thanks!

Regarding weird capitalisation, Italian Wikipedia does indeed have "eSwatini", as does [treccani](https://www.treccani.it/enciclopedia/regno-di-eswatini).

Looking at Dutch Wikipedia, the title of the article is still "Swaziland", so arguably the "country" name should be that (and Eswatini only mentioned in the country info).  However, given that we've had Eswatini/eSwatini until now, it seems silly to switch back to the old name, since it'd be confusing for existing users.

reinpeter:
Yes, there is always a nice 'debate' on what to use, the older (badly?) transl(iter)ated (kolonial?) names or the (newer) official names that are used by the countries itself...  Don't know if wikipedia is always right ;)  (Sortof similar to the Peking/Beijing question)
In this case, our government agrees on the same priority, so seems ok like this :) 
https://www.nederlandwereldwijd.nl/reisadvies/eswatini

PR #588, opened 2022-12-18
Small changes / additions (T&C Eilanden, Malediven, Oostzee)

reinpeter:
1) Fix a taipoh, 'eianden' should be 'eilanden'. https://nl.wikipedia.org/wiki/Turks-_en_Caicoseilanden

2) Malediven is more common in Dutch. https://nl.wikipedia.org/wiki/Malediven 

3) Baltic sea is also a very common alternative name. https://nl.wikipedia.org/wiki/Oostzee


axelboc (approved):
Thanks!

aplaice (approved):

PR #596, opened 2023-03-20
italy and hungary flags are similar via rotation

lepras:


aplaice:
Thanks for the PR!

However, are Hungary and Italy really two countries whose flags you confuse or is it just something that you think should be added?

<hr/>

(The idea behind the "flag similarity" field is to provide information which might help learners who confuse the flags in question because of the similarity, not simply list all flags that resemble the given flag.)

(Also, currently, rotation isn't [among the criteria](https://github.com/anki-geo/ultimate-geography/blob/master/CONTRIBUTING.md#flag-similarity-field) that we consider when determining "similarity", but that could be changed.)

lepras:
"Major differences (M): (1) same colours in different positions (e.g. two swapped, three `rotated`)"

https://dailynewshungary.com/wp-content/uploads/2016/02/italy.jpg

Just exposing yourself to the flags and their countries, is how I approached the cramming and I genuinely get them confused most of the times. A 90 degree **exact** rotation doesn't really change the perception to the naked eye if you know what I mean.

axelboc:
Fair enough. Though, how many other cases are there? If it's a couple, no worries, but if we have to add 10 more similarities because of this change in the guidelines, it might be a bit much.

lepras:
I have only done european flags till now, found just one such case.

aplaice:
Sorry for the late reply!

> "Major differences (M): (1) same colours in different positions (e.g. two swapped, three *rotated*)"

By "rotated" colours [we mean](https://github.com/anki-geo/ultimate-geography/wiki/Flag-similarities) a [cyclic permutation](https://mathworld.wolfram.com/CyclicPermutation.html) of the colours (e.g. {red, white, black} → {white, black, red}), rather than a geometrical rotation. This might not be clear from the wording, though. :) (Should we change it?)

> Though, how many other cases are there?

The only other cases that I can think of (without doing an extensive survey) are the (France, Luxembourg, Netherlands) group and the (Peru (assuming lack of coat of arms), Austria, Latvia (possibly)) one. (All other possible pairs/groups have an emblem in at least one of the flags, which would disqualify the pairing/grouping.) (I haven't looked at the colour differences, here (or in the Hungary, Italy pair either, for that matter).)

_If_ we were to do this, should a geometrical rotation be a critical or a major difference?

Also, how would we deal with geometrical differences (width etc.)? (The flags of Hungary and Italy have different geometries (a ratio of 1:2 and 2:3), but when the bands are rotated, that difference might not matter, anyway.

lepras:
> By "rotated" colours [we mean](https://github.com/anki-geo/ultimate-geography/wiki/Flag-similarities) a [cyclic permutation](https://mathworld.wolfram.com/CyclicPermutation.html) of the colours (e.g. {red, white, black} → {white, black, red}), rather than a geometrical rotation. This might not be clear from the wording, though. :) (Should we change it?)
> 

I think you should mention cyclic permuted rather than rotated.

> _If_ we were to do this, should a geometrical rotation be a critical or a major difference?

After reading their differences and criteria for similarity, critical seems more apt.

axelboc:
Getting back to this, sorry for the radio silence 😅 

While I don't deny that one might confuse these two flags, I still think that their two greens and their two ratios differ significantly enough that considering the two flags as similar is a bit of a stretch. We don't want hints for more similar flags to get lost in the noise, especially since there are already a lot of hints.

I'm not against adding "90° colour rotation" as a similarity criteria, but I would be more inclined to categorise it as a **critcial difference** in order to qualify pairs of such flags as similar only if they don't have any major/minor differences (thus disqualifying Hungary/Italy, sorry 😁).

aplaice:
Closing since I don't think that it's likely we'll be merging this.  Sorry @lepras!

(I'm not opposed to adding 90° rotations for flag differences, as such, but since I don't confuse these categories of flags, I'm definitely not the right person to spearhead this.)

PR #599, opened 2023-04-16
[SV] Change from Vitryssland to Belarus 

melissis:
Sweden's Ministry for Foreign Affairs officially started using the current name in November 2019, and media has switched as well. Some people still use the old name, of course.


axelboc (approved):
Thanks! Wikipedia for reference https://sv.wikipedia.org/wiki/Belarus

PR #607, opened 2023-08-13
Fix: re-lock dependencies to fix `ValueError` during install

samm81:
running `pipenv install` would get a `ValueError: "Requirement" object has no field "use_pep517"`, the internet suggested that running `pipenv lock` would fix the issue - and it did


aplaice (approved):
Thanks!

I can't reproduce the initial error (even after updating pipenv), but I do get the same Pipfile.lock as you after re-locking (`pipenv lock`). (No idea what's the underlying cause, but it (almost) never hurts to update dependencies :).)

PR #617, opened 2023-12-26
Add tags UG::South_Africa, UG::North_Africa, UG::Central_Africa.

natsume42:


horwitz:
Are these well-defined? i.e., are there undisputed definitions of what countries fall under these terms?

What about issues with South Africa the country vs. a region called South Africa?

natsume42:
I used [Wikipedia](https://en.wikipedia.org/wiki/United_Nations_geoscheme_for_Africa) as a reference.
This is also in line with `UG::East_Africa` and `UG::West_Africa`, which are already used in this deck.

horwitz:
Thanks—and good point re Wikipedia (and UN). And thanks for doing the work here

Perhaps it'd also be good to update the tags to match Wikipediia (and [The UN source](https://unstats.un.org/unsd/methodology/m49/#geo-regions))—i.e., using Northern, Eastern, Western, Southern, and Middle Africa (instead of North, East, West, South, and Central Africa)? (This would solve the South Africa (country) / South Africa (region) issue, make the terminology more internally consistent, and, of course, match Wikipedia (and the UN).)

aplaice:
Thanks @natsume42 for the contribution, but I'm not sure if we want these tags.

> Are these well-defined? i.e., are there undisputed definitions of what countries fall under these terms?

No, AFAICT not really. There are many different definitions, sometimes using slightly different names (South Africa vs. Southern Africa or Middle Africa vs. Central Africa) but often using the same name.

(I believe that North Africa is well defined, but it has only a couple of members, so I'm not sure if it's worth the tag.)

See [these](https://github.com/anki-geo/ultimate-geography/issues/244#issuecomment-619439698) [comments](https://github.com/anki-geo/ultimate-geography/pull/304#issue-610440143) (and associated threads) for some details on the issues with the categorisations. (East and West Africa were chosen because they're _relatively_ well-defined and substantive.)

<hr/>

@natsume42 please do feel free to disagree and push back, though!

natsume42:
@aplaice: Sure.
I will close the PR, as I feel that this change is not worth further discussions.

axelboc:
Just a quick tip for anyone landing here: it's possible to get the equivalent of "Northern + Middle + Southern Africa" by creating a filtered deck with the following query: `tag:UG::Africa -tag:UG::East_Africa -tag:UG::West_Africa`. This gives 23 notes (versus 21 for  East Africa and 16 for West Africa).

PR #620, opened 2023-12-31
Update Artsakh and Azerbaijan

### LINKED ISSUE #611: Remove Republic of Artsakh
The Republic of Artsakh will cease to exist no later than January 1, 2024 (most of its population has fled already):
https://www.reuters.com/world/asia-pacific/nagorno-karabakh-republic-will-cease-exist-jan-1-2024-nagorno-karabakh-2023-09-28/

aplaice:
* Remove Republic of Artsakh

It no longer controls any of its claimed territory.

Fix #611.

~As of 2023-12-31 it is still listed at:~

~https://en.wikipedia.org/wiki/List_of_sovereign_states#Other_states~

~so according to our contributions guidelines we might (?) still want to wait(?). (Somebody has already opened a discussion in the [relevant talk page](https://en.wikipedia.org/wiki/Talk:List_of_sovereign_states#Artsakh_-_gone_now?).)~ It's been now removed from the list, so we can probably safely merge.

It also appears that the government of Artsakh will not be dissolved as previously declared but will instead be a government-in-exile (but IMO that is not, in itself sufficient ground to keep the note).

* Update map of Azerbaijan

It now controls Artsakh, so it doesn't make sense to have it marked as disputed territory (or: more precisely it's disputed territory between an internationally-recognised sovereign state and a stateless government-in-exile that is/was not internationally recognised, so probably closest to the status of ~Tibet, which we do not mark as disputed).


aplaice:
Artsakh's been now removed from [the list](https://en.wikipedia.org/wiki/List_of_sovereign_states#Other_states), so we can probably safely merge.

axelboc (approved):
I think the `sources.csv` file also needs updating – then, it's good to go.

aplaice:
> I think the sources.csv file also needs updating – then, it's good to go.

Oops! Thanks for noticing!

PR #624, opened 2024-02-12
Add simple instructions for upgrading from v3.3- APKG

### LINKED ISSUE #594: "The front of this card is blank" on some cards after updating
Hello and thanks for your work on the project. I've learned a lot through it and I love it ❤️ 

This week I have upgraded from the apkg format to v5.1 and it went smoothly, other than a few cards coming up during review as "The front of this card is blank" with no further info.

The cards involved are (at least) Northern Mariana Islands, Guernsey, Cayman Islands.

I'm a bit lost as to what happened, and how it can be fixed. I understand people updating from apkg is going to be increasingly rare, but it might be worth a mention in the upgrade instructions.

aplaice:
This is deliberately kept short, to avoid too much info in the README.md that's likely only applicable to a few people (rather old version).

Fix #594.


axelboc (approved):
Awesome, thanks :blush:

PR #630, opened 2024-02-22
Added Swedish translation for deck header.

melissis:
Untested.


melissis:
I added a bonus pair of line breaks that I don't think was in the English version.

melissis:
I just created a new file on github.com, copied English text and pasted. Is it in UTF-8 now or do I need to do something more?

src/headers/desc_sv.html
```diff
@@ -0,0 +1,17 @@
+<a href="https://github.com/anki-geo/ultimate-geography/"><b>FULLSTÄNDIG BESKRIVNING</b></a> | <a href="https://github.com/anki-geo/ultimate-geography/releases"><b>UTGIVNINGSANMÄRKNINGAR</b></a> | <a href="https://github.com/anki-geo/ultimate-geography/blob/master/CONTRIBUTING.md"><b>BIDRA</b></a>
+
+<b>Ultimate Geography v5.2</b> innehåller:
+
+- världens <a href="https://sv.wikipedia.org/wiki/Lista_%C3%B6ver_sj%C3%A4lvst%C3%A4ndiga_stater"><b>205 självständiga stater</b></a> (820 kort)
```
aplaice, line 5: I _think_ that it's best to use the English Wikipedia link here, because that's what we rely on for the listing. For instance, we include (disputed) sovereign states like Taiwan, Palestine or South Ossetia.

aplaice (approved):
Thanks!

The file is indeed UTF-8!

<hr/>

In order for the deck header to be used by CrowdAnki, you also need to change `recipes/source_to_anki.yaml`, basically copying the Spanish version (or the Chinese version):

https://github.com/anki-geo/ultimate-geography/blob/dc3eed70da1f98735e5c9e266aa77d8f04961714/recipes/source_to_anki.yaml#L34C1-L39C40

(but with a new randomly generated UUID) and then specifying the new header for the two Swedish decks [here](https://github.com/anki-geo/ultimate-geography/blob/dc3eed70da1f98735e5c9e266aa77d8f04961714/recipes/source_to_anki.yaml#L404) and [here](https://github.com/anki-geo/ultimate-geography/blob/dc3eed70da1f98735e5c9e266aa77d8f04961714/recipes/source_to_anki.yaml#L495), again similarly to how it's done for Spanish/Chinese.

(Or if you prefer I can do it.)

melissis:
> (but with a new randomly generated UUID)

How do I get a UUID? ; v ;

aplaice:
> How do I get a UUID? ; v ;

If you're on a Linux system, then `uuidgen` will generate one. If you have access to a python runtime, then `import uuid; str(uuid.uuid4())` will also do the job. Otherwise, the website https://www.uuidgenerator.net/ seems to do the job.

axelboc (approved):

PR #647, opened 2024-08-17
Add brief link to Wiki page describing card mappings to switch decks

aplaice:
TBH I'm not sure if this is needed for now.

We can link to the Wiki page wherever we advertise the new interactive deck, and we can more prominently link the page in the README when we adjust it to better describe the interactive deck once it's more "stable".


axelboc (approved):
This is great, I think!

PR #651, opened 2024-08-24
Anki v24.06+ back card side automatic swap fix

helitopia:
- See [the discussion comment](https://github.com/anki-geo/ultimate-geography/discussions/649#discussioncomment-10433459) for the details

### DIFF
--- a/src/media/experimental_assets/_ug-interactive_map_init.js
+++ b/src/media/experimental_assets/_ug-interactive_map_init.js
@@ -1,3 +1,8 @@
+/**
+ * Interactive map initialization logic declaration.
+ * Note that due to Anki Desktop web view being persistent
+ * for card reviews IIFE is used to separate the namespaces
+ */
 (function () {
   const mapConfig = getMapConfig();
   const commonConfig = mapConfig.commonConfig;
--- a/src/media/experimental_assets/_ug-interactive_map_init.js
+++ b/src/media/experimental_assets/_ug-interactive_map_init.js
@@ -175,8 +175,8 @@
   }
 
   /**
-   * Trigger "Enter" key press event. Note that Anki &lt; 24.06
-   * uses `code` property and Anki &gt;= 24.06 - `key` property
+   * Trigger "Enter" key press event. Note that Anki < 24.06
+   * uses `code` property and Anki >= 24.06 - `key` property
    * to query pressed key, so both properties must be present
    */
   function dispatchEnterEvent() {
--- a/src/media/experimental_assets/_ug-interactive_map_init.js
+++ b/src/media/experimental_assets/_ug-interactive_map_init.js
@@ -165,8 +165,18 @@
     if (typeof AnkiDroidJS !== "undefined") {
       showAnswer();
     } else {
-      commonElements.hiddenTextarea.dispatchEvent(new KeyboardEvent("keypress", {code: "Enter"}));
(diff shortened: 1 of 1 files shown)


helitopia:
Force pushed to resolve merge conflict as forgot to sync master branch in fork with the upstream.

axelboc (approved):
Maybe a little comment to explain that `code` is for backwards compatibility with Anki <= x.y?

helitopia:
> Maybe a little comment to explain that code is for backwards compatibility with Anki <= x.y?

I hesitated when writing commit message, as it is already mentioned in the discussion:

> I tried to fix it by setting both properties (for it to work on both pre-24.06 and post- versions): new KeyboardEvent("keypress", {code: "Enter", key: "Enter"}). It worked, so I will open PR to resolve the issue (@aplaice @axelboc or should I open issue first?).

And since I am linking the discussion thread, thought that it would be redundant.

axelboc:
I'm all for writing readable code and limiting the number of comments, but this is a case where a very short comment can be super helpful in the long term.

It can be tedious to find discussions linked to a piece of code: one has to "blame" the specific line to find which commit last modified it then go to the corresponding PR and then to the related discussion. It becomes a lot more difficult when that piece of code gets moved to a different file or gets changed multiple times...

In 5 years' time, if it stops working or if someone stumbles upon this code and thinks that `code` is redundant and wants to remove it, it'd be good to inform them right away that they should first check whether the version of Anki it targets is still in use or not.

aplaice:
Thanks very much for diagnosing and fixing this!

This is nitpicking, but I'd also go with a comment in the code. (My preference here: comment > git commit message > link.) (Having the info available in `git blame` is nice, but IMO this is one of these "weird" things that ideally should just be commented in the code.)

helitopia:
It is not a problem, completely agree that a shortcut for reasoning would be helpful. I guess I can then move IIFE reasoning from wiki into code as well.

helitopia:
Separated into two commits, one for the card swap and the other - IIFE docs.

Not sure why, but I am somewhat triggered by regular comments in code. Probably due to having prior experience with the project that compensated for shitty code with comments everywhere :). Hence, refactored into separate function with proper JSDoc comment above it.

src/media/experimental_assets/_ug-interactive_map_init.js
```diff
@@ -165,10 +170,20 @@
...
     } else {
-      commonElements.hiddenTextarea.dispatchEvent(new KeyboardEvent("keypress", {code: "Enter"}));
+      dispatchEnterEvent()
     }
   }
 
+  /**
+   * Trigger "Enter" key press event. Note that Anki &lt; 24.06
+   * uses `code` property and Anki &gt;= 24.06 - `key` property
+   * to query pressed key, so both properties must be present
```
axelboc, line 180:
```suggestion
   * Trigger "Enter" key press event. Note that Anki < 24.06
   * uses `code` property and Anki >= 24.06 - `key` property
   * to query pressed key, so both properties must be present
```

axelboc (approved):
Thanks a lot!

src/media/experimental_assets/_ug-interactive_map_init.js
helitopia: Should I correct it? I see the PR is merged, not sure why though.

PR #654, opened 2024-10-29
Restore alternative names for the Sea of Galilee

kusunosetoru:
Due to the change by #582, country_info.csv contains two lines for the Sea of Galilee.
Because of this, preexisting alternative names listed in the first line are no longer contained
in the current decks in the affected languages.

Merging the Dutch alternative name to the first one fixes this.


axelboc (approved):
Well spotted, thanks!

aplaice:
Thanks for catching this!

(And sorry for having let this in!)

<details><summary>It seems to be the only duplicate</summary>

```bash
for i in src/data/*.csv
do
    csvtool cols 1 "$i" | sort | uniq -c | grep -v 1
done
```

returns only:

```
      2 Sea of Galilee
```

</details>

PR #659, opened 2024-11-30
Fix two errors in the danish translation of country.csv

marcmtk:
Fix two errors in the danish translation of country.csv, one type in mikronesien and one error where South america (Sydamerika) was labelled middle america (Mellemamerika).

This is my first pull request to this repo, and i'm a little unsure after reading CONTRIBUTING.md if editing the csv is sufficient. Please let me know if there's additional steps to be taken.


axelboc (approved):
Thanks! That's all there's to it 💯

PR #669, opened 2025-01-23
Bump python version

aplaice:
3.11 picked because it's the latest version I have on my system.  (Latest to minimise the number of future upgrades.)

(The `build/` output is the same as before, when running `pipenv run brain_brew run recipes/source_to_anki.yaml`.)


aplaice:
@axelboc should we merge? I don't think we can easily resolve the broken action without upgrading our python version (well without doing something crazy like downloading python 3.7 ourselves, but that's not really optimal :D) (as you had noted).

I don't have strong opinions on the particular python version, though.

axelboc (approved):
Sorry, meant to approve earlier. All good!

PR #627, opened 2024-02-17
CONTRIBUTING.md: state flag vs. civil flag

horwitz:
address issue #111

PR #674, opened 2025-02-10
Update based on recent(-ish) changes to article names on Wikipedia

aplaice:
Fix #627.

* Rename East Timor to Timor-Leste

    Change in English only (none of the other Wikipedias have changed the 
article name).  (However, also change the parenthetical name in Chinese.)

    Note that for the Wikimedia sources of the map/flag, the upstream name 
hasn't been changed, so the links in sources.csv are unchanged.

* Fix typo in Portuguese version of Timor-Leste

    (Lack of hyphen!)

* Update spelling of Sanaa in English and Czech

   English Wikipedia updated the main name in 2019: 
   https://en.wikipedia.org/w/index.php?title=Sanaa&oldid=928109035

   Czech Wikipedia updated it on 2023-01-01.

   Danish Wikipedia has Sanaá as the main spelling, but "Danske Sprognævn"
   (main Danish source) still has Sana.

   The old spelling has been added to the capital info to avoid confusion (we
   don't want people who learned the old spelling to think that what they
   learned is wrong when their deck is updated).

*   Clarify Contributing guidelines for minor spelling variations


CONTRIBUTING.md
```diff
@@ -253,7 +253,7 @@ If multiple capitals are listed in a country's infobox, the following guidelines
 
 As explained in the previous section, this field is typically used for countries with multiple capitals, to clarify the role and/or status of each capital, or to explain succinctly why the country has multiple capitals. In such case, the content of the field should be consistent across languages.
 
-This field may also be used to provide alternative capital names and spellings, which may be language-specific.
+This field may also be used to provide alternative capital names and spellings, which may be language-specific.  Minor spelling differences should usually be ignored, unless the main spelling was explicitly changed, on Wikipedia (for instance Kiev to Kyiv or Sana'a to Sanaa). In these cases, the old spelling is included mainly to avoid confusion for previous users of the deck. See [#672][ref672] and [#381][ref381].
```
axelboc, line 256:
```suggestion
This field may also be used to provide alternative capital names and spellings, which may be language-specific.  Minor spelling differences should usually be ignored, unless the main spelling was explicitly changed on Wikipedia (for instance Kiev to Kyiv or Sana'a to Sanaa). In these cases, the old spelling is included mainly to avoid confusion for previous users of the deck. See [#672][ref672] and [#381][ref381].
```

axelboc (approved):
💯

PR #682, opened 2025-03-20
Update flag of Syria to that of the revolution

### LINKED ISSUE #660: Syrian flag?
Since 8 December (when the Assad regime fell) the transitional government of Syria has used the green-white-black horizontal stripes with three red stars; see https://en.wikipedia.org/wiki/Flag_of_Syria

aplaice:
Fix #660.

I think enough time has passed that it's unlikely that there'll be back-and-forth.

I _think_ that we could include this in v5.3.

(Also: fix copy-paste error in ZH-TW. (New Zealand copied into previous row.))


axelboc (approved):
Good call! And good thinking removing the flag similarities.

aplaice:
I _think_ that the new Syrian flag is not "similar" (according to our criteria) to any other flags (the closest is the UAE), so I don't think we need to add any new flag similarity.  However, I only briefly skimmed through all the current flags, so I might have missed something.

PR #684, opened 2025-04-06
Add script for zipping decks for releases/sharing

aplaice:
This was originally designed for the automatic uploading of files to the unofficial "everfresh" release in my personal fork, and still keeps the parts of the code relevant only to that purpose.

I'm not at all sure whether this needs to be included in the main repo (I'm happy to keep it in my personal files), but I thought it might be worth it as simple, but useful maintenance code.

(_If_ we are to include it, it might be preferable to move it into some subfolder or to remove the old github-CI-specific parts of the code.  (OTOH if we do figure out a way to run attaching files to releases in GitHub actions in a safe way, this code might turn out to be useful in the main repo, as well.))

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -397,6 +397,7 @@ Content changes, such as adding a note, replacing an image, or translating the d
 1. Add each folder in the `build` directory to a separate ZIP archive named as follows:
    - `Ultimate Geography [EN]` ==> `Ultimate_Geography_v[x.y]_EN.zip`.
    - `Ultimate Geography [EN] [Extended]` ==> `Ultimate_Geography_v[x.y]_EN_EXTENDED.zip`.
+   This can be done with `pipenv run zip`.
 1. On GitHub, create a new **release** named after the version number.
 1. Draft the release notes, making sure to add a link to the upgrade steps in the `README` and/or [in the wiki](https://github.com/anki-geo/ultimate-geography/wiki/Upgrade-instructions).
 1. Attach all the ZIP files and save the draft release notes.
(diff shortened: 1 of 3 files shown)


axelboc:
I had written a script like this for Node a long while back (code below for reference). The only difference I think is that it was reading language codes directly from the names of the generated deck folders in the `build` directory.

<details>
<summary><code>anki-archiver</code></summary>

```js
const path = require('path');
const fs = require('fs');
const { zip } = require('@papb/zip');

const ROOT = path.normalize('/.../ultimate-geography');
const BUILD = path.join(ROOT, 'build');

const descPath = path.join(ROOT, 'src/headers/desc.html');
const desc = fs.readFileSync(descPath, { encoding: 'utf-8' });
const [, version] = desc.match(/Ultimate Geography (v[0-9]+\.[0-9]+)/) || [];

const buildEntries = fs.readdirSync(BUILD, { withFileTypes: true });
const deckFolders = buildEntries.filter(entry => entry.isDirectory()).map(entry => entry.name);

(async () => {
  for (let name of deckFolders) {
    const target = name
      .replace(/ /g, '_')
      .replace(/\[([A-Z]{2})\]/, '$1')
      .replace('[Extended]', 'EXTENDED')
      .replace(/^(Ultimate_Geography)_(.*)$/, (_, ug, rest) => (
        `${ug}_${version}_${rest}.zip`
      ));

    await zip(path.join(BUILD, name), path.join(BUILD, target));
  }
})();
```

</details>

I think it's definitely worth having a script like this in the repo!

It'd be great to make it a first-class citizen, though - i.e. to write it in Python (for cross-platform support) and integrate it with Pipenv (dependency for zipping in `Pipfile` + `pipenv run zip` script).

axelboc (approved):
I can confirm it works beautifully on my end! 🥇

axelboc:
Oh can you please just add running this script as a step in the maintainer's guide?

PR #705, opened 2025-08-31
Update flag of Hungary based on Wikimedia

### LINKED ISSUE #704: Update flag of Hungary
In AUG we now have the ["19:28, 4 June 2018" version from Wikimedia](https://commons.wikimedia.org/w/index.php?title=File:Flag_of_Hungary.svg&offset=&limit=50). The current version on Wikipedia/Wikimedia is different and has been in place since 4 January 2023 (unlikely to be a temporary edit war version). As far as I can tell (based on the archived links to pantone and the Hungarian Standards Institution (?)) the colours are the "official" ones. (There was an edit war on Wikimedia in ~2022 based on the question of whether the "official" colours or those used in practice are the "correct" ones, but it appears the "official" colours side won.)

I believe we should update the flag to the current Wikimedia version.

This should not change any of the [past flag similarity decisions related to Hungary](https://github.com/anki-geo/ultimate-geography/wiki/Flag-similarities) (none were dependent on the precise shades of red/green of the Hungarian flag).

aplaice:
Fix #704.

AFAIU this is not a change due to external (legal) changes, but based on what Wikipedia decided is the correct choice of flag colours.  (The new colours are the legally/officially-specified colours.)

The new, updated flag has been on Wikipedia since 4 January 2023, so it's relatively unlikely to be a temporary "edit war" version.


axelboc (approved):
Very subtle, thanks!

PR #707, opened 2025-08-31
Expand on the viewBox attribute in CONTRIBUTING.md docs

aplaice:
I always have doubts about what I should change when updating flags, and (I believe) always end up doing what's described in this commit.

The weasel-word "usually" is in case there are more unusual situations such as the width/height being missing or being specified as a percentage.


axelboc (approved):
That covers most cases I think 💯

PR #712, opened 2025-09-17
Add flag similarity script

aplaice:
This is meant as an aide to determine whether two flags are similar or not.  It does not replace visual comparison, but it does replace the extracting colours from the SVGs and inputting them into the ΔE* spreadsheet.

Testing (not systematically) it seems that the script correctly regenerates the colour difference values we had obtained with the spreadsheet:
   https://github.com/anki-geo/ultimate-geography/wiki/Flag-similarities

(Approximately: we use a slightly different conversion factor from the reference spreadsheet, but I believe that our factor is the correct one.)

Writing the script turned out considerably more time-consuming than expected at the outset, due to increasing numbers of edge-cases...

The script is messy and I'm not sure whether anyone other than myself will find it useful, but given that it's been written, having it is IMO better than not having it...

<hr/>

In the cases where there are few colours, it works well:

image
image

but when there are many, it becomes messy:

image


axelboc (approved):
Wow that's quite impressive. Definitely better than a spreadsheet... No notes :grinning:

PR #714, opened 2025-09-26
German country info fix

radioproof:
The wording is somewhat inconsistent in the german translation. Here are some proposed fixes.

1. For claimed countries use the following wording consistently: [...] anerkannter Staat, von [...] beansprucht.
2. For British and French overseas territories change two notes for each to match the other notes.
3. Make the wording for autonomous regions use the same grammar. Instead of having some say 'Autonome Region von Portugal' and others say 'Autonome Region Italiens'. Use the latter version consistently.
4. Switch around the sentences for Aaland to match Turks and Caicos note logic.
5. Align translation for Taiwan and French Guiana with english version.


aplaice (approved):
Thanks very much!  Consistency is very much desired!  To the extent that I can tell, this looks great!

(My German is pretty terrible, so I'm not capable of accurately determining whether these are stylistically better.  (e.g. "von Marokko beansprucht" does seem to sound better, but I don't trust my assessment, so I need to rely on your judgement that this is the correct direction of applying consistency.))

AFAICT (under a very cursory look) the changes also better align with the wording in German Wikipedia (e.g. [Autonome Gemeinschaften Spaniens](https://de.wikipedia.org/wiki/Autonome_Gemeinschaften_Spaniens) or britisches Überseegebiet in the [Cayman Islands article](https://de.wikipedia.org/wiki/Cayman_Islands).), which (all other things equal) we usually seek.

PR #723, opened 2025-11-23
Update some flags

aplaice:
This is the first tranche of flags updated with the (unmerged) update flags script (going alphabetically).  None of the changes are major.

For each flag I've reviewed what's the actual change, the modification history on Wikimedia and whether the justification actually made sense (usually not delving too deeply).  The modification history is unfortunately usually not very clean: 1. frequently they're daisy chains of "Reverted to version as of {date_x}", where the change on {date_x} is in turn a reversion to date_y etc., 2. the actual modification messages are almost always rather terse.

(I've not included two flags where the change was clearly only due to svgo — it might make sense to add them once we decide on whether (and how) to version svgo.  I've also not included Australia and Canada where the source (the wikimedia file) of the flag displayed in the relevant country and flag articles has changed.  (I haven't looked closely, but one seem to be due to disputes about whether to use the official pantone colours (converted to rgb) or directly the official rgb colours and the other about whether to adhere to the "normal" Union Jack colours or not.))

### Motivation of changes

Most of the changes are due to Wikipedia/Wikimedia better following existing guidelines (or due to guidelines actually appearing — many flag designs are underspecified in the original laws) or improving on the pantone->RGB conversion.

Three are due to actual external changes.  Azerbaijan's change is due to a modified law.  Aruba's _probably_ also is.  Austria's is due to changed guidelines (there's no "official" colours, but the Ministry of Defense switched recommendations from Pantone 032 C to Pantone 186 C in 2018).

### Procedures going forward

I'm not sure how best to record the information for each change.  I've been committing each flag separately, as it makes it easier to keep track (this is pretty boring, so I've been doing it in small batches), but having a separate commit for each of the (in total) ~200 slightly changed flags feels ludicrous.  OTOH keeping some of this information, especially where there's an external motivation, might be useful.  _Probably_ the best approach is to squash all the commits, heavily abbreviating the info.

Going forward I'll probably review the changes in less detail — check whether the change is (visually) minor, that the last modification on wikimedia was more than two (?) years ago and only dig deeper if it was more recent.

### DIFF
--- a/src/media/flags/ug-flag-colombia.svg
+++ b/src/media/flags/ug-flag-colombia.svg
@@ -1,1 +1,1 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="375" height="250" viewBox="0 0 6 4"><path fill="#CE1126" d="M0 0h6v4H0z"/><path fill="#003893" d="M0 0h6v3H0z"/><path fill="#FCD116" d="M0 0h6v2H0z"/></svg>
+<svg xmlns="http://www.w3.org/2000/svg" width="375" height="250" viewBox="0 0 900 600"><path fill="#ffcd00" d="M0 0h900v600H0z"/><path fill="#003087" d="M0 300h900v300H0z"/><path fill="#c8102e" d="M0 450h900v150H0z"/></svg>


axelboc (approved):
Brilliant! There are more visible changes than I thought there would be. :clap:

Whatever works for you but squashing into a single commit like "Update flags from A to C" would make indeed sense I think.

PR #731, opened 2026-02-04
Update more flags (C-Z) using script

aplaice:
This is the everything from this "pass".  I'll do another one later (see below).

I'll squash all the commits when merging.  (For convenience, I had verified the flag changes (looking at the visual diff + skimming the flag version history and digging deeper if something seemed dubious) one by one, with one country per commit, and a brief description of the change/motivation if warranted.)

### Officially changed flags

For easy reference (extracting from the git log) here are the cases where the flag was ~legally changed (or in some cases clarified where it was previously ambiguous):

* France (Macron reverted to the pre-1974 colour; we had discussed this in #531, but hadn't applied it)
* Honduras (shade of blue officially changed)
* Mauritius (2022 National Symbols Act)
* Oman (in 2004 (!), the flag's proportions were changed to 4:7)
* Panama (new 2017 law specifying colours)
* Thailand (Colours officially standardised in 2017!  They even provided CIELAB values!)
* Vatican City (2023)

### Other notes

Most of the remainder is:

* better conversion of pantone colours
* better adherence to guidelines (or better choice of guidelines/sources where there's nothing definitive and/or the "definitive" sources conflict)
* tweaks to details/revectorisation/optimisation

<hr/>

I haven't applied changes where there was no visual diff (unfortunately svgo is not deterministic — I haven't investigated, but it seems that it sometimes switches the order of attributes).  I'll add this into the next version of the update flag script.

<hr/>

### Wikimedia changes not applied

Below are the cases (I kept notes mostly for my own reference) where I did not apply changes from Wikimedia.  These are often when there's an ongoing edit war or I'm not sufficiently convinced that the new version is strictly better (more accurate) than our current version.  (Our considerations are different than Wikipedia's/Wikimedia's: their goal is purely to have, at all times, as accurate an image as possible; we seek an accurate image, but we also want to avoid unnecessary changes, since that is likely to confuse users.)

I'll do a second pass at some point soon-ish (probably several weeks) and review them.

* Australia
* Canada (I hadn't kept notes, yet, but I believe both Australia and Canada had conflicts about Pantone colours and official RGB colours which don't match each other.)
* Egypt (Yellow colour!) (too recent and IMO wrong (based on real-life images))
* Eswatini (We changed the tassels to black in 2022; Wikimedia has changed back to purple.  I'm not convinced either way and there's a long-standing edit-war.  Some government sources support black others blue/purple.  Hence, I don't want to switch back, for now.)
guam (too recent (2025-12 !!!))
* Iran (recent (2025-06) and IMO not well justified (while the _penultimate_ change appears to have been well-sourced, explained and discussed on the talk page))
* Jamaica (recent (2025-09); there have been several independent changes and I believe most of them are correct, but I'm not entirely convinced by the final one)
* Kazakhstan (I think that the new Wikimedia flag is an improvement over our current one, but there's a long-ish (though not very recent) edit-war over the precise shades.)
* Lebanon (recent (2024-12 for substantial change)) (probably OK)
* Malta (long-ish existing change (from 2023), but IMO **wildly** wrong.  Based on a [random-ish specification](https://www.gov.mt/en/About%20Malta/Documents/Maltese%20Islands/National%20Flag%20Graphical%20Specifications.pdf) which doesn't mention the gray of Saint George's cross (but which, on the larger images, has the interior depicted as gray!!!), some Wikipedian has decided that the interior of Saint George's cross should be white).
* Mexico (the flag is 150 kB (!) and AFAICT the visual change is very, very tiny — the colour of the fruits of the cactus on which the eagle sits — and the last change was rather recent (2025-06))
* Montenegro (IMO the penultimate version is correct, while the ultimate one is wrong.  I've brought it up on the talk page: https://commons.wikimedia.org/wiki/File_talk:Flag_of_Montenegro.svg )
* Nicaragua (The changes are tiny (very slight change of the shade of the mountains in the emblem), all of the changelog comments since 2014 say that they're only optimizing/reducing code (i.e. any minor changes aren't actually intended) and I'm too lazy to redo the gaussian blur. :D)
* Niue (The version of svgo I was using mangled the flag.  (Removing stars by converting scale("0.000...x") to scale("0").))
* Norway (There was a consensus on the talk page, a long time ago (2021) and no substantive change since then, but looking at photos of real-world examples of the flag, I am very much not convinced...
* Palestine (Ongoing edit war (last change 2025-11).)
* Papua New Guinea (Last change in 2025-12 (I don't think there's anything controversial, but I'd prefer to wait a while.))
* Paraguay (I don't see absolutely any difference (when displaying one after the other), though there is a very tiny visual diff; I don't want to blur the flag :D)
* Puerto Rico (I'm not convinced — the latest change to the shade of blue seemed to be random.)
* South Africa (The colours aren't officially specified, but there was an svg on an official site and the new colours don't match it.  I'm not convinced that using the colours from the government svg is the right approach, but I'm also not convinced by the new wikimedia colours.  (NB our current colours are different from both.)
https://web.archive.org/web/20201001213827/https://www.gov.za/sites/all/themes/custom/eco_omega/images/flag-south-africa.svg
https://web.archive.org/web/20210625013331/https://www.gov.za/documents/constitution-republic-south-africa-1996-schedule-1-national-flag )
* South Korea (I believe that in 2011, according to a new law, the colours were adjusted to lighter shades — leading to the new wikimedia version.  OTOH someone (19:18, 5 August 2017) claims that government sources still support a darker shade and discussed this in the talk page; however, they seem to have been ignored (other than a vigorous edit war in 2024).  I haven't looked closely whether these darker shades are the same as the pre-2011 ones and whether there's actually any support for them (and how they compare to our versions).)
* South Ossetia (Active-ish edit war (2024-12) and unresolved talk page discussion.)
*Spain (I'm not convinced by the justification for the latest change.  (Procedure was sort-of followed — bringing up topic in talk page, but the reasoning was AFAICT absent.))
* Sri Lanka (AFAICT There have been two separate and conflicting updates of the flag: 1. "29 December 2020" using various construction sheets and 2. "7 April 2022" using SLS 1:2020 (flag size number 5).  Proponents of **1** say that **2** has "too many errors"; while proponents of **2** say that it's still a government standard.  I'm not convinced either way.)
* Transnistria (Ongoing edit war.)
* United States (Ongoing discussion about colours in flag article talk page.)
* Uzbekistan (Ongoing edit war.)
* Zambia (Ongoing edit war (large change to shade of green))


axelboc (approved):
Wow that's a lot! 💪

PR #735, opened 2026-05-22
Add Hebrew deck translation

mikecrawford99:
## Summary

- Add Hebrew (`HE`) as a supported language across note data, note model metadata, deck recipes, deck descriptions, and documentation.
- Add reusable RTL support to the generated note model templates via `__TEXT_DIRECTION__` and `__IS_RIGHT_TO_LEFT__`.
- Add Hebrew translations for countries, capitals, country/capital info, capital hints, and flag similarity text.
- Add Hebrew deck builds for standard, extended, and experimental variants.
- Add a static Hebrew deck description and update existing deck descriptions/README language lists to include Hebrew.

## Implementation

- `src/note_models/translations.csv` now includes text direction metadata for all languages and Hebrew template substitutions.
- Textual Anki fields now receive `is_right_to_left` metadata where applicable.
- Card templates wrap rendered card content in `dir="__TEXT_DIRECTION__"` so Hebrew cards render right-to-left while existing languages remain left-to-right.
- `recipes/source_to_anki.yaml` and `recipes/source_to_anki_[experimental].yaml` now build Hebrew decks and include `guid:he`.
- `src/headers/desc_he.html` is tracked as a static localized deck description, matching the pattern used by the other deck descriptions.
- `CONTRIBUTING.md` documents the Hebrew translation source convention.


aplaice (approved):
Thanks very much! Unfortunately, I can't review the actual translations, but the "plumbing" all looks fine! (Good catch on the previously unreplaced `Hint` in the template! Thanks!)

For the card text direction, we could probably also set it via css, rather than the `div dir`s but it'd require templating of the CSS, so this is probably simpler (and probably more flexible (?), in case we ever need mixed text direction in cards).

mikecrawford99:
@aplaice:
> Thanks very much! Unfortunately, I can't review the actual translations, but the "plumbing" all looks fine! (Good catch on the previously unreplaced `Hint` in the template! Thanks!)
> 
> For the card text direction, we could probably also set it via css, rather than the `div dir`s but it'd require templating of the CSS, so this is probably simpler (and probably more flexible (?), in case we ever need mixed text direction in cards).

Awesome, thanks! Yeah, I figured the divs were the safest bet for flexibility. 
Do you know around when you'll be able to merge this in?

aplaice:
>  Do you know around when you'll be able to merge this in?

We usually wait a while in case anyone else has additional feedback (especially from the core team), but given that none of the core team speaks Hebrew and given that there's a chance of significant structural changes in the near future, I'm merging this now.  (If there's need we can always edit later.)

May I add you to the [translators](https://github.com/anki-geo/ultimate-geography/wiki/Translators) list?

mikecrawford99:
> > Do you know around when you'll be able to merge this in?
> 
> We usually wait a while in case anyone else has additional feedback (especially from the core team), but given that none of the core team speaks Hebrew and given that there's a chance of significant structural changes in the near future, I'm merging this now. (If there's need we can always edit later.)
> 
> May I add you to the [translators](https://github.com/anki-geo/ultimate-geography/wiki/Translators) list?

Sure, no problem.

PR #740, opened 2026-07-20
Add Malacca strait map

### LINKED ISSUE #739: Strait of Malacca
Would you consider adding the Strait of Malacca (https://en.wikipedia.org/wiki/Strait_of_Malacca)? Geopolitically, this is an important one (https://en.wikipedia.org/wiki/Malacca_dilemma).

Clement-Jean:
I open this PR to add the Malacca strait. As of now, I created the map but I don't know how to add the card. If you could guide me that would be great.

see: #739


aplaice:
Wow! Thanks!

I think that following our conventions we should use the name "Strait of Malacca" (the title of the corresponding [English Wikipedia article](https://en.wikipedia.org/wiki/Strait_of_Malacca)), rather than Malacca Strait (even though the latter is also frequently used), which would also cause the name of the map image to change (`ug-map-strait_of_malacca.png`). (If you disagree, please say so!)

> add the card

You need to:

1. Append the "Strait of Malacca" to `src/data/country.csv`, ideally with translations in the corresponding columns (this is slightly tedious... — typically I open the Wikipedia articles in the corresponding languages linked from the English article (in the left sidebar if you use Wikipedia's old interface or at the top if you use the new one)).

    (We could technically automate this (resolving Wikipedia interlanguage links), but we'd still need to manually check that none of the other-language articles are disambiguation pages (as occasionally happens).)

    If you don't want to do this, then feel free to just add the translations in any other languages you know and leave the rest.

2. Append it to `src/data/main.csv`, with no flag (obviously), an HTML img element for the map, and a `UG::Oceans+Seas` tag.
3. Ideally/optionally, if there are any alternative names of the strait in any of the relevant languages, these would need to go at the end of `src/data/country_info.csv`. English at least doesn't have any other major alternative names (Strait of Malacca vs. Malacca Strait doesn't count according to our guidelines), so it's possible that you don't need to modify `country_info.csv`.

    (I usually glance at the article intro of each language when doing step **1** to see if there are any other names, at least in the languages that I can (at least sort of) parse, but it's not a significant issue if we miss some — if there are important alternative names, then users will eventually point them out.)

4. Append "Strait of Malacca" to `src/data/guid.csv` (without filling in the guids themselves) and run `pipenv run build` to generate the guids (or leave it to us to run).

<hr/>

If you've taken any other existing map from Wikimedia as a base for the map, please add it as a source in `sources.csv`, mentioning any modifications in the Modifications column. Alternatively, you can upload an SVG version of the map to Wikimedia and then add that as the source.

<hr/>

Thanks!

Clement-Jean:
hope this is good, let me know if there are any changes that need to be made

sources.csv
```diff
@@ -545,3 +545,4 @@ ug-map-yemen.png,https://commons.wikimedia.org/wiki/File:Yemen_in_its_region.svg
 ug-map-zambia.png,https://commons.wikimedia.org/wiki/File:Zambia_in_its_region.svg,CC BY-SA 3.0,
 ug-map-zanzibar.png,https://commons.wikimedia.org/wiki/File:Zanzibar_in_its_region_(special_marker).svg,CC BY-SA 3.0,
 ug-map-zimbabwe.png,https://commons.wikimedia.org/wiki/File:Zimbabwe_in_its_region.svg,CC BY-SA 3.0,
+ug-map-strait_of_malacca,https://commons.wikimedia.org/wiki/File:Sumatra_in_its_region.svg,CC BY-SA 4.0,
```
aplaice: Unlike the files in `src/data/*.csv`, `sources.csv` is alphabetically sorted, so could you please move this row upwards to maintain the sorting? (Sorry for not stating this beforehand and sorry for the inconsistency.)

Also, could you add a "Modifications" comment, maybe something like "highlight changed"?

src/data/country.csv
```diff
@@ -318,3 +318,4 @@ Gulf of Guinea,Golf von Guinea,Golfo de Guinea,Golfe de Guinée,Guineabukta,Guin
 North Sea,Nordsee,Mar del Norte,Mer du Nord,Nordsjøen,Severní moře,Северное море,Noordzee,Nordsjön,Mar do Norte,北海(North Sea),北海 (North Sea),Morze Północne,Mare del Nord,Nordsøen,הים הצפוני
 Sea of Okhotsk,Ochotskisches Meer,Mar de Ojotsk,Mer d'Okhotsk,Okhotskhavet,Ochotské moře,Охотское море,Zee van Ochotsk,Ochotska havet,Mar de Okhotsk,鄂霍次克海(Sea of Okhotsk),鄂霍次克海 (Sea of Okhotsk),Morze Ochockie,Mare di Ochotsk,Det Okhotske Hav,ים אוחוצק
 Timor Sea,Timorsee,Mar de Timor,Mer de Timor,Timorsjøen,Timorské moře,Тиморское море,Timorzee,Timorsjön,Mar de Timor,帝汶海(Timor Sea),帝汶海 (Timor Sea),Morze Timor,Mar di Timor,Timorhavet,ים טימור
+Strait of Malacca,Straße von Malakka,Estrecho de Malaca,Détroit de Malacca,Malakkastredet,Malacký průliv,Малаккский пролив,Straat Malakka,Malackasundet,Estreito de Malaca,马六甲海峡,麻六甲海峽,Cieśnina Malakka,Stretto di Malacca,Malaccastrædet,מצר מלאקה
```
aplaice: I think that 麻六甲海峽 should be 馬六甲海峽:


```suggestion
Strait of Malacca,Straße von Malakka,Estrecho de Malaca,Détroit de Malacca,Malakkastredet,Malacký průliv,Малаккский пролив,Straat Malakka,Malackasundet,Estreito de Malaca,马六甲海峡,馬六甲海峽,Cieśnina Malakka,Stretto di Malacca,Malaccastrædet,מצר מלאקה
```

aplaice (approved):
Thanks very much for such a fast update!

Other than a couple of nitpicks, this looks great!

Clement-Jean:
Should be done. Thank you for the guidance and approval.

aplaice:
Thanks again!

PR #749, opened 2026-08-03
Rename Nauru to Naoero

patrickmcanuff:
Nauru officially changed its name to the Republic of Naoero. Parliament passed the constitutional amendment in May 2026; President Adeang confirmed on 29 July that the planned referendum was dropped ("Naoero is not a new name seeking acceptance by the people"), and the UN has listed the country under the new name since a government request in June. Australia, New Zealand, the US and China have updated their foreign-ministry sites.

Sources: [RNZ](https://www.rnz.co.nz/news/pacific/836342/nauru-officially-changes-name-to-naoero-after-thoughtful-deliberation-bypassing-referendum) · [Al Jazeera](https://www.aljazeera.com/news/2026/7/31/pacific-island-nation-nauru-officially-changes-name-to-naoero) · [ABC/AP](https://abcnews.com/International/wireStory/nauru-worlds-3rd-smallest-country-naoero-135318134)

### What changed

The country name is the join key for every source CSV, so the rename is applied atomically across `main`, `guid`, `country`, `country_info`, `capital`, `capital_info` and `flag_similarity` — a partial rename would silently drop fields at the join.

- **GUIDs deliberately unchanged.** Note identity is the GUID, not the country name, so existing users keep their review scheduling. Verified: the built note's GUID is byte-identical to master's.
- **Curaçao's `flag_similarity` cross-reference** named Nauru in 12 languages and is updated too — easy to miss, since it lives on a different card.
- **Media renamed** to `ug-flag-naoero.svg` / `ug-map-naoero.png` per the `ug-<type>-<country_name>` convention in CONTRIBUTING.md:276,332, with the matching `sources.csv` attribution rows. The Wikimedia source URLs keep their original filenames.
- **New `country_info` row**, `Known as Nauru until 2026.`, following the existing Eswatini/Swaziland precedent.

### Deliberately not changed

- **ISO stays `NR`.** ISO 3166 has not been updated — alpha-2 is still `NR`, alpha-3 `NRU`. The `NRO` code in press coverage isn't reflected in the standard yet, so this is worth revisiting later.
- **Non-Latin-script names** (`ru`, `zh`, `zh-tw`, `he`) keep the old name. Transliterating "Naoero" is a call for native speakers, and a wrong transliteration is worse than a stale-but-correct old one. This matches how Eswatini still carries `es: Suazilandia` and `sv: Swaziland`. **Native speakers of those four languages: input welcome.** The same applies to the empty `cs` cell in the new `country_info` row.
- **`_ug-world.js`** still has `name:"Nauru"`. That vendored jsvectormap data is matched by region code, which is empty for this country, so the label is unused — and I'd rather not touch a 1.3 MB third-party blob in a content PR.

### Verification

`pipenv run build` succeeds. In the built deck: 2 notes reference Naoero (the country note and Curaçao's similarity note), 0 notes still read `Nauru`, and the GUID matches source. Also checked that `main.csv` and `sources.csv` remain correctly sorted, and that the CSV writer round-trips byte-identically so the diff stays minimal after #747.


aplaice:
Thanks for picking this up!

According to our [contribution guidelines](https://github.com/anki-geo/ultimate-geography/blob/master/CONTRIBUTING.md#country-field) we use the names that the corresponding Wikipedias use and mention any alternative names in `country_info.csv`.  Hence, based on the current state on Wikipedia, for English, we would have `country`: `Nauru`, `country_info`: `Officially Naoero` (or maybe `Officially Republic of Naoero`?).

However, given that there's an article move/rename (Nauru->Naoero) in discussion now, I think it's best to wait a while (~6 months?) for the various Wikipedias to settle (they might settle on different choices, depending on how frequently each version is used in the given language), to avoid churn!

patrickmcanuff:
Makes sense, thanks for the pointer — I'd missed that the Wikipedia article title is the source of truth here. Happy to go either way: I can slim this down to just the `country_info` addition (`Officially Republic of Naoero`) for now, or close it and revisit once the article discussions settle. Whichever you prefer!

aplaice:
Thanks for the reply!

I think revisiting this in a while (and maybe opening an issue in the mean-time as a reminder (?)) is the simplest and cleanest approach.

Doing the "slimmed down" approach properly (including all languages) would IMO be a waste of effort, especially since Wikipedia is in a state-of-flux (e.g. two days' ago German Wikipedia had migrated to Naoero as the article name, but they appear to have since reverted).

patrickmcanuff:
That all makes sense — agreed the slimmed-down version isn't worth the churn while things are in flux. I've opened #751 as a reminder to revisit once the Wikipedia titles settle, and I'll close this in the meantime. The branch will still be around when the time comes. Thanks for the guidance!

PR #750, opened 2026-08-04
Fix minor (country|capital)_info issues

aplaice:
(Caught when reviewing unrelated PR.)

* Fix Portuguese country_info for Isle of Man

* Ensure that all country and capital info fields end in full stop

  Both Chinese (zh/zh-tw) columns skipped in country info since I'm not confident about when the Chinese full stop (。) should/should not be used.


axelboc (approved):
LGTM 😊

PR #756, opened 2026-09-05
fix: correct Chinese translation of Dominica

### LINKED ISSUE #754: The Chinese translation of country Dominica is incorrect
Thank you to all the developers and translators. I became interested in world geography because of the World Cup, and this project helps me a lot. I have found a Chinese translation issue:

According to Wikipedia: <https://zh.wikipedia.org/zh-cn/多米尼克>
the Chinese translation of Dominica should be "多米尼克", instead of "多米尼加", to distinguish it from Dominican republic (多米尼加共和国).

Both simplified Chinese (zh) and Traditional Chinese (zh-tw) should be modified.

zerovip:
Fixes #754


aplaice (approved):
Thanks! AFAICT (i.e. comparing with Wikipedia and [Wiktionary](https://en.wiktionary.org/wiki/%E5%A4%9A%E7%B1%B3%E5%B0%BC%E5%85%8B)), this looks great! (But feedback from our translators would still be useful, in case I've missed something or misunderstood!)

